PR #48, opened 2015-05-03
Add position info to entries in `keys`.

izaakschroeder:
Generating URLs from parameters becomes infinitely easier when positional information is provided about those parameters. The only changes to output are two new properties: `position` and `length`; they represent the start and length of a variable, respectively. For the string `/:foo/:bar` the two entries would have positions 1 and 6, and lengths 4 and 4. This change is API compatible and shouldn't break any existing use of the library.

There exists another library to generate URLs from parameters, https://github.com/krakenjs/reverend, but its approach is a significantly bigger hack. This feature would allow for a much more streamlined and efficient generation algorithm.

Specs have been updated to match.

### DIFF
--- a/index.js
+++ b/index.js
@@ -19,7 +19,7 @@ var PATH_REGEXP = new RegExp([
   //
   // "/:test(\\d+)?" => ["/", "test", "\d+", undefined, "?"]
   // "/route(\\d+)" => [undefined, undefined, undefined, "\d+", undefined]
-  '([\\/.])?(?:\\:(\\w+)(?:\\(((?:\\\\.|[^)])*)\\))?|\\(((?:\\\\.|[^)])*)\\))([+*?])?',
+  '([\\/.])?((?:\\:(\\w+)(?:\\(((?:\\\\.|[^)])*)\\))?|\\(((?:\\\\.|[^)])*)\\))([+*?])?)',
   // Match regexp special characters that are always escaped.
   '([.+*?=^!:${}()[\\]|\\/])'
 ].join('|'), 'g')
@@ -110,7 +110,7 @@ function arrayToRegexp (path, keys, options) {
 function replacePath (path, keys) {
   var index = 0
 
-  function replace (_, escaped, prefix, key, capture, group, suffix, escape) {
+  function replace (_, escaped, prefix, all, key, capture, group, suffix, escape, i) {
     if (escaped) {
       return escaped
     }
@@ -126,7 +126,9 @@ function replacePath (path, keys) {
       name: key || index++,
       delimiter: prefix || '/',
       optional: optional,
-      repeat: repeat
+      repeat: repeat,
+      offset: i + (prefix ? prefix.length : 0),
+      length: all.length
     })
 
     prefix = prefix ? ('\\' + prefix) : ''
--- a/test.js
+++ b/test.js
@@ -65,14 +65,14 @@ var TESTS = [
   ['/test/', [], '/test//route', ['/test'], { end: false }],
   [
     '/:test',
-    [{ name: 'test', delimiter: '/', optional: false, repeat: false }],
+    [{ name: 'test', delimiter: '/', optional: false, repeat: false, offset: 1, length: 5 }],
     '/route',
     ['/route', 'route'],
     { end: false }
   ],
   [
     '/:test/',
-    [{ name: 'test', delimiter: '/', optional: false, repeat: false }],
+    [{ name: 'test', delimiter: '/', optional: false, repeat: false, offset: 1, length: 5 }],
     '/route',
     ['/route', 'route'],
     { end: false }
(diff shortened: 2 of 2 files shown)


blakeembrey:
Damn, thanks for the work! I took another approach a bit ago (not PR'd) where I added a new property called `tokens` to the RegExp and that has the keys and string tokens in sequential order. What do you think the advantage of doing this is over the tokens array? Would either be just as easy to work with?

izaakschroeder:
Tokens is fine too! They're naturally a little more heavy-weight (storing bunch of strings instead of just offsets), but they're also easier to work with.

The build function with tokens amounts to:

```javascript
import _ from 'lodash';
_.chain(tokens)
  .map(token => token.type === 'variable' ? lookup(token.value) : token.value)
  .join('')
  .value();
```

Generating the tokens from the ranges amounts to something like:

```javascript
import _ from 'lodash';
_.chain(re.keys)
  .reduce((prev, cur) => {
    const [last, out] = prev;
    return [cur.end, out.concat(url.substr(last, cur.start), url.substr(cur.start, cur.end))];
  }, [0,[]])
  .value()
```

blakeembrey:
I really like this solution. I definitely think there's a case for creating paths from the strings, though it can only be generated from strings and not the other currently supported types. I think it'd be good to add a little 10 liner to transform strings/regexp (from string) instances into to paths. There's some basic rules to take into account, but all the info is there now so it sounds great :+1:

Just let me know if you are taking that on and I can review your next PR :smile:

index.js
```diff
@@ -126,7 +126,9 @@ function replacePath (path, keys) {
       name: key || index++,
       delimiter: prefix || '/',
       optional: optional,
-      repeat: repeat
+      repeat: repeat,
+      offset: i + (prefix ? prefix.length : 0),
```
blakeembrey: I'm seeing what I can do about supporting the path logic and wondering why you decided to do this? I would want the prefix to be included with all because when the group is optional the prefix shouldn't be output.

izaakschroeder: The prefix is already included in the key entry as `delimiter`, so it can be added if it's needed?

izaakschroeder: I have some code already for this, I can push it up for you if you'd like. It's pretty prototypy but if you wanna see I'll post it.

blakeembrey: I think it should be, otherwise you need to minus 1 every time. This also allows you to get length with just `_.length` (rename that to `all` or something useful though). If you want, go ahead. I was starting to write it too and it's pretty neat.

blakeembrey: I'll PR my code too for you to review. There's a little more that needs to be added to support this, but I think it was a nifty start. I added a "pattern" property (we can validate the user input against it) and the "prefix" property (it's like delimiter, but sometimes there's no prefix and we need that knowledge).

blakeembrey: Ok, I'm thinking tokens might be a better approach now. Since we need to remove escape characters from the output and I think it makes sense to provide as much help to the user as possible. Thoughts?

blakeembrey: @izaakschroeder Hey! Did you want to check out https://github.com/pillarjs/path-to-regexp/pull/50? I pulled everything into a `parser` function that anyone can use and exposed it, then re-implemented the `toRegexp` functionality using it. Takes the extra bits and bops into account too, so there's no more hacks needed.

izaakschroeder: Checking now...

blakeembrey: @izaakschroeder I spent some time wiring it up with the path logic (https://github.com/pillarjs/path-to-regexp/pull/50/files#diff-168726dbe96b3ce427e7fedce31bb0bcR98). Let me know what you think. It'll give you an idea of the various cases too.

izaakschroeder: Nice @blakeembrey! Sorry for the delay in checking this out been busy couple of days here. You think it's worth attaching keys _and_ tokens now? Since they're always computed anyway and will contain references to the same objects it will be cheapish. So in addition to `foo.keys` there is now also `foo.tokens`. In an ideal world you eventually deprecate `keys` since everything needed is in `tokens` anyway.

izaakschroeder: As post-merge review thought, you might consider replacing some common utility functions with their well-tested lodash versions. e.g. `function omit (keys, src) {` can just be `_.omit`. Since it's just in testing it won't change the weight of the library (or if you're extra concerned you can just import the functions you need since modular lodash exists).

blakeembrey: Yes, I know. You're right. The omit there is actually because I was too lazy to update all the tests immediately, I'll be removing that later. I don't think there's anything I actually need `lodash` for, but you're right that build size is a valid concern (not that it matters in tests, like you mentioned).

blakeembrey: > You think it's worth attaching keys and tokens now?

If you check if out, you'll see that `parse` returns an array of keys and strings, then the `toRegexp` functionality turns that into a regexp and just an array of keys. Since there's actually no need for the strings at that moment. I do think it's worth returning both in the parse step since there's additional logic that implementors can easily get wrong otherwise (for example, un-escaping characters). I think it's valuable to provide the most functionality we can right out of the box, since it doesn't add much overhead here (we're already matching for escaped characters).

PR #50, opened 2015-05-06
Expose path parser

blakeembrey:
With this PR we expose the `parse` method to consumers. This can be used to get the segments of the path, as well as the sanitised strings (currently we're just removing escape characters from things the user may have escaped from us). From this, it's trivial to construct a path builder and the `toRegexp` functionality will use the parser from now on.

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -1,6 +1,6 @@
 # Path-to-RegExp
 
-Turn an Express-style path string such as `/user/:name` into a regular expression.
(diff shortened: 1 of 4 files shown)


blakeembrey:
Just run a few speed tests and didn't find any compilation slowdowns (in fact, it improves, if anything). This is great, even though compilation speed isn't the most important thing since it should only happen once.

PR #53, opened 2015-05-12
Expose tokens to regexp/function methods

### LINKED ISSUE #52: Ability to convert tokens to regex.
The #parse method returns an array of tokens, it would be cool if we could pass in those tokens to get a regex again.

It would be useful to merge multiple parsed paths.

I could do a PR if anyone else things this is a good idea.

blakeembrey:
Closes #52


rill-js:
Looks good. Thanks for the quick response!

Issue #37, opened 2014-08-26
Path wildcards


recidive:
I was using pathRegexp() from from Express utils, but since this was removed in 4, I had to migrate to path-to-regexp. I was using it to match an URL only (to get the context for a page) not actually using parameters.

The problem is that /some-path/\* URLs stopped working and I need to change it across all the application to use /some-path/(.*).

Since this is an application to be used also by non programmers, /some-path/\* would be more straightforward.

Do you have any plans to bring back /some-path/\* at some point?

andineck:
I ran into this problem as well, and it took me a while to find the reason for this unexpected behaviour, since page.js docs are not reflecting the current behaviour.

I just checked express.js deps again:     
"path-to-regexp": "0.1.3",

I really think the current version where wildcard does not work anymore as it does in express.js is very confusing for the user. regex is not everyone's favourite and it makes things not very readable. 

I can understand, the ambiguity in the implementation, but I would definitively prefer a consistent (as it has been for a while) and simple use of the module, over a simple implementation of the module itself.

blakeembrey:
@andineck I understand the concern, but there shouldn't be a case where you need to use `*`. Could you give a code snippet where you use it? In 99% of cases, you should be using `.use` in Express.js which is also a slight optimisation since you don't need to do path matching. For everything else, you can use a regular expression which, I think, is much clearer about the intention. I understand it doesn't make things very readable, but I also don't think `*` is any more readable than `/:path*` - the latter even gives you a key name. Also, the asterisk behaviour was largely undefined. For instance, it just expanded to `(.*)` which made key indexes are to keep in `0.1.x`, and it can cause unexpected behaviour. For example, `/user/:path*/foo` will match `/user/123/foo` but `/user/*` will match the same except it can match `/user//foo` since it's just a zero or more matching group. The expansion syntax is a little more complex in that it'll actually take into account what the path looks like.

andineck:
I'm not saying there is a use case that could not be solved with `(.*)` instead of  `*`.

I am using express.Router on the server side, and page.js in the browser.
Now page.js switched to the new version, express sticks with 0.1.3 (strict).

my path is `/upload/*` and it is matching e.g. '/upload/img/small'
I am not saying this use case could be rewritten to something like `/upload/(.*)` or `/upload/:path*`,
but I also don't think these are more readable than `/upload/*`

The `*` was not documented in the best way, but it is documented after all:

Express:
[...]

PR #54, opened 2015-05-12
Support asterisk as an unnamed matching group

blakeembrey:
Supports `/*` once again.

### DIFF
--- a/index.js
+++ b/index.js
@@ -21,9 +21,10 @@ var PATH_REGEXP = new RegExp([
   // Match Express-style parameters and un-named parameters with a prefix
   // and optional suffixes. Matches appear as:
   //
-  // "/:test(\\d+)?" => ["/", "test", "\d+", undefined, "?"]
-  // "/route(\\d+)" => [undefined, undefined, undefined, "\d+", undefined]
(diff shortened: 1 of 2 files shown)


dougwilson:
Rather than making a special exception for /* would it be desirable just to have the rule of when a path segment contains only the character * ? i.e. for /foo/*/bar or something? Idk, just throwing out comments now that i subscribed to this repo.

blakeembrey:
@dougwilson Right now there's no logic matching path segments, but I guess it could be done. In this PR, it's being treated exactly as if `*` was `(.*)`. So both `/foo*` and `/foo(.*)` would act the same. Could you elaborate a little more though? Maybe there's some advantage doing one way or over the other.

andineck:
@blakeembrey thx. it works as you mentioned in #37 as far as I can see:
https://gist.github.com/andineck/f863a38d699976830de9

same behaviour as the version 0.1.3 with unnamed parameter with asterisk.
with named parameters with asterisk there is a difference as mentioned, that with this PR it matches everything starting with the asterisk, not only the path segment as in 0.1.3.

PR #57, opened 2015-06-20
fix using asterisk inside param regexp

dougwilson:
This is my attempt at fixing https://github.com/strongloop/express/issues/2495 in the 0.1.x series :)

### DIFF
--- a/index.js
+++ b/index.js
@@ -32,6 +32,7 @@ function pathtoRegexp(path, keys, options) {
   var end = options.end !== false;
   var flags = options.sensitive ? '' : 'i';
   var extraOffset = 0;
+  var starOffset = 0;
   var keysOffset = keys.length;
   var i = 0;
   var name = 0;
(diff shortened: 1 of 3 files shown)


blakeembrey:
Doesn't this break backward compatibility with `0.1.x`? I know there's a few intentional uses with it, which is why I've avoided touching it.

dougwilson:
Sure, I guess in the technical sense, but I think I've heard enough from users that it is very unexpected that the stuff `:foo(<here>)` gets transformed in any way.

blakeembrey:
Yes, I agree. But that's the gist of `1.x`. Couldn't you just upgrade to `1.x` then?

dougwilson:
> Couldn't you just upgrade to 1.x then?

No, because that breaks paths like `/*`...

blakeembrey:
@dougwilson Not since `1.2.x`. See https://github.com/pillarjs/path-to-regexp#asterisk.

dougwilson:
Ah I see. So you're saying if I were to try 1.2.x in Express right now, all our tests would just pass?

blakeembrey:
@dougwilson No guarantees, since there's a lot of other fixes, but I can give it a try :+1:

dougwilson:
Darn it, we have an example with the path `'/files/:file(*)'`, so there goes the neighborhood :) I really wished I could fix this :)

blakeembrey:
Just beat me to it. Yeah, that's the only major difference between `0.1` and `1.x` now. Well, plus there's a lot of new stuff in `1.x`.

dougwilson:
Cool. I can't wait to get 5.0 out with 1.x. Thanks for keeping me honest :)

PR #68, opened 2016-01-11
browser support added

erayarslan:


blakeembrey:
I'm not sure how much was changed, but it looks like you just copied and pasted the code from `index.js`, changing the dependency to be inline? If you're intention is to create a bundle without dependencies, this is not enough for "browser support" as it's polluting the global scope and all logic is duplicated. If you want a browser bundle, I recommend using Browserify or Webpack.

PR #72, opened 2016-03-02
Add TypeScript `.d.ts` file for type checking

blakeembrey:


dougwilson:
This is super awesome! What I wound wonder, especially if we start adding this to other modules, is I under if there is a way to use TypeScript to valid the bindings file against the JavaScript file? Is that even possible through some kind of command line?

blakeembrey:
@dougwilson I spent some time thinking about it from last year, and yes, there is. I was considering making a more complex toolchain, but an easy way is to just use the compiler. We can either compile a `.ts` file and run the emitted JavaScript (E.g. type check, then a runtime check) or we could do it all in one step using something like `ts-node`. I can append it to this PR, which would you like to see or use?

dougwilson:
I'm not certain which one makes the most sense :) What I'm thinking is that either there could be a `npm run-script` command that would say yes/no or some way to make it be a `mocha` test, so like `it('should contain valid TypeScript definitions', ...)` or something.

blakeembrey:
Hmm, ok. I was thinking you'd just have part of the test suite as literally a TypeScript file and that would enable the type checking on what you're doing. Are you think more along that there should be a way to load the JavaScript file and the definition file and "overlay" them to check if they're valid? I don't think TypeScript has anything like that, at least not right now, but the advantage of writing it against the actual test suite would be that we continuously test the defined options parameters, etc.

blakeembrey:
@dougwilson I committed a version using `ts-node` as part of the test suite, what do you think?

PR #85, opened 2016-06-23
Fix TypeScript 2 error

calebmer:
This syntax is not supported in TypeScript 2

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -56,4 +56,4 @@ declare namespace pathToRegexp {
   export type PathFunction = (data?: Object, options?: PathFunctionOptions) => string;
 }
 
-export = pathToRegexp;
+export default pathToRegexp;


blakeembrey:
Yes, it is.

blakeembrey:
There's a huge difference between `export =` and `export default` - they are _not_ compatible in any way. The syntax you tried to switch it to does not work for this module, and it broke the build. I'd recommend reading up on the differences via the TypeScript docs.

blakeembrey:
This is probably a good place to start: https://github.com/Microsoft/TypeScript-Handbook/blob/master/pages/Modules.md

calebmer:
Thanks, I apologize. I just learned the difference and was just about to close the PR. Sorry for the distraction and the inapropriate OP.

PR #86, opened 2016-07-15
Fix inconsistency between "exec" and "compile" when parameters with modifiers contain slashes

TiuSh:
Reading through the tests I found usecases where the behavior of "exec" and "compile" seems to be inconsistent : 

```js
var re = pathToRegexp('/:test*');
re.exec('/some/basic/route');
// ['/some/basic/route', 'some/basic/route']
var toPath = pathToRegexp.compile('/:test*')
toPath({ test: 'some/basic/route' });
// /some%2Fbasic%2Froute
toPath({ test: ['some', 'basic', 'route'] });
// /some/basic/route
```

From what I understand from the docs these usecases should look like this :

```js
var re = pathToRegexp('/:test*');
re.exec('/some/basic/route');
// ['/some/basic/route', 'some/basic/route']
var toPath = pathToRegexp.compile('/:test*')
toPath({ test: 'some/basic/route' });
// /some/basic/route
toPath({ test: ['some', 'basic', 'route'] });
// /some/basic/route
```

What do you think ?

### DIFF
--- a/index.js
+++ b/index.js
@@ -177,6 +177,10 @@ function tokensToFunction (tokens) {
         }
       }
 
+      if (typeof value === 'string' && token.repeat) {
+        value = value.match(new RegExp('([^' + escapeString(token.delimiter) + ']+)', 'g'))
+      }
+
       if (isarray(value)) {
         if (!token.repeat) {
           throw new TypeError('Expected "' + token.name + '" to not repeat, but received `' + JSON.stringify(value) + '`')
--- a/test.ts
+++ b/test.ts
@@ -739,6 +739,7 @@ var TESTS: Test[] = [
     [
       [{}, ''],
       [{ test: 'foobar' }, '/foobar'],
+      [{ test: 'foo/bar' }, '/foo/bar'],
       [{ test: ['foo', 'bar'] }, '/foo/bar']
     ]
   ],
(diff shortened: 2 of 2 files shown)


blakeembrey:
This currently works as intended. It's meant to escape the string input.

TiuSh:
I understand that it was intended as it was quite explicit from the tests :)

But if a pattern `/:test*`, tested on the path `/some/basic/route` resolves the parameter `test` as `some/basic/route`, I would expect to be able to generate back the path `/some/basic/route` using the same pattern (i.e. `/:test*`) and parameter (i.e. `some/basic/route`), as it work for any other type of pattern (even `/*`).

Am I missing something ?

blakeembrey:
It only works for `*` since that's a match everything already. For `/:test*` you need to pass an array of values. It doesn't make sense to allow plain strings as then there's no way to know what's encoded and what's not encoded and breaks the useful of having a compile feature if it's not deterministic.

blakeembrey:
Just a note, but you know you could just use `String#split`? At least it looks like that's what you're trying to do anyway.

TiuSh:
I'm still convinced that when using a pattern like `/:test*` or `/:test+`, it should be my responsibility to properly escape the `test` parameter. Otherwise I could use the array format to make sure that each segment is escaped by the lib. But you have the last word anyway ;)

As for `String#split`, I preferred `String#match` for a much wider browser support.

PR #88, opened 2016-08-23
Remove isarray, use native Array.isArray()

Chyzwar:


blakeembrey:
No, thanks. If you submit a PR, make sure make the minimum number of required changes (E.g. no `console.log` comments). With that said, this change is a breaking change for anyone using the current version and I don't really see the need for that right now.

Chyzwar:
Sorry for console.log but how it is breaking change ? isArray is from chrome version 5

blakeembrey:
Because you're removing functionality. Removing functionality would be a breaking change. What if there's a downstream module that guarantees IE8 support? However unlikely, it's still a breaking change for anyone using this module.

DavidTPate:
The change is unnecessary anyways the `isarray` module uses `Array.isArray` if it is available, otherwise it implements an alternative. You can see that [here](https://github.com/juliangruber/isarray/blob/2f9b4daa7cd8002383987dbfc7863fde6563d272/index.js#L3)

blakeembrey:
Correct 👍 It is definitely preferred to use `Array.isArray` without a dependency though, but as that's how this is implemented for reasons mentioned above no need to change it 😄

PR #94, opened 2016-10-02
Enable a default delimiter for parsing behaviour

blakeembrey:
@DylanPiercey - this might help you with handling hostnames. Let me know 😄

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -31,6 +31,7 @@ var pathToRegexp = require('path-to-regexp')
   - **sensitive** When `true` the route will be case sensitive. (default: `false`)
   - **strict** When `false` the trailing slash is optional. (default: `false`)
   - **end** When `false` the path will match at the beginning. (default: `true`)
+  - **delimiter** Set the default delimiter for repeat parameters. (default: `'/'`)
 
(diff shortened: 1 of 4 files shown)


DylanPiercey:
@blakeembrey looks sweet to me, thanks for looking into working with hostnames.

Currently the lib works fine for me with hostnames and I am having a little trouble following exactly whats changed in terms of functionality. Could you elaborate what this commit has fixed or changed at a high level?

blakeembrey:
Yes, sorry. I enabled `delimiter` to be passed in as a `parse` option. Before it assumed the default delimiter to always be `/`. The only place this would come up is in `:foo+.com` where it'd assume `a/b/c` instead of `a.b.c` because of the default delimiter character.

DylanPiercey:
Ah cool, thats sweet. I saw the delimiter option but wasn't exactly sure what it was doing.

I have never personally had a situation where I wanted one or more domains in a group but this is sweet functionality to add and will definitely help people (like me) who might have guessed that would work before :p.

Would love to see this merged so I can update Rill :D.

Thanks for your work!

PR #100, opened 2017-03-28
add umd

yurypaleev:


wesleytodd:
Personally I am a 👎 to this.  I would suggest making a wrapper module, like `umd-path-to-regexp`, which wraps this up in it.

yurypaleev:
@wesleytodd, and how can I use such module in the webpage?

wesleytodd:
Depends on your tools and setup.  If you published a new node module which just required `path-to-regexp` and wrapped it in the UMD format, you could then include it in your browserify or webpack build just like any other module.  The UMD wrapper would take care of exposing the global variable for you if you needed it in other places on your website.

yurypaleev:
What is the use of wrapping? If I need to use browserify or webpack to get the code, eventually. This additional code will solve the problem on many fronts.

wesleytodd:
If you use webpack or browserify then this module needs no changes for you to be able to use it.  My assumption was you wanted UMD so you could get a global variable export or use it with require.js.

Either way, UMD is not something the core express/pillar team wants to support/add to modules.

yurypaleev:
I would like to use it in the browser, keeping this module especially does not require anything to implement it.

wesleytodd:
You can use it in the browser via webpack or browserify without any changes.  And it does have a cost to the maintainers and users to include UMD support directly in the module when it is not needed.

yurypaleev:
Okay

Issue #73, opened 2016-03-24
Do we really need isArray as a dependency ?


sakthiiv:
After all the npm left-pad problem, do we require isArray ? Its just a single line of code .. `toString.call(arr) == '[object Array]'`

blakeembrey:
Yes.

blakeembrey:
Removing this breaks backward compatibility at this point, you're welcome to inline the code if you need to in a PR.

wtgtybhertgeghgtwtg:
What browsers would be supported by a `2.0`?

blakeembrey:
@wtgtybhertgeghgtwtg Well, there is no 2.0 yet, but most likely every browser that it does today and it'd just require polyfills for certain features instead.

wtgtybhertgeghgtwtg:
Is a `2.0` planned?

PR #104, opened 2017-04-18
Remove isarray dependency

frenzzy:
I think it is better to don't include shims and polyfills to the library code and only add notes about used ES5/ES6/ES7 features to the readme file. It allows to avoid code duplication and decrease final bundle size for most of projects.

`Array.isArray()` compatibility:

[Desktop browsers](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/isArray#Browser_compatibility)

Feature | Chrome | Firefox (Gecko) | Internet Explorer | Opera | Safari
-- | -- | -- | -- | -- | --
Basic support | 5 | 4.0 (2.0) | 9 | 10.5 | 5

[Mobile browsers](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/isArray#Browser_compatibility)

Feature | Android | Chrome for Android | Firefox Mobile (Gecko) | IE Mobile | Opera Mobile | Safari Mobile
-- | -- | -- | -- | -- | -- | --
Basic support | (Yes) | (Yes) | 4.0 (2.0) | (Yes) | (Yes) | (Yes)

[Node.js](http://node.green/#ES2015-subclassing-Array-is-subclassable-Array-isArray-support)

7.9.0 | 7.5.0 | 6.10.2 | 6.4.0 | 5.12.0 | 4.8.2 | 0.12.18 | 0.10.48
-- | -- | -- | -- | -- | -- | -- | --
Yes | Yes | Yes | Yes | Yes | Yes | Error | Error

This is a **BREAKING CHANGE** which requires increasing of the major version by [semver](http://semver.org/).
Maybe it is a good time to release `2.0`? :)

Refs: #88, #73

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -221,6 +221,14 @@ Path-To-RegExp exposes the two functions used internally that accept an array of
 * `pattern` The RegExp used to match this token (`string`)
 * `asterisk` Indicates the token is an `*` match (`boolean`)
 
+## Browser Support
+
+Path-To-RegExp supports all popular browsers, including Internet Explorer 9 and above.
+
+For compatibility with older browsers that don't support ES5 methods you may need to include
+[`Array.isArray()` polyfill](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/isArray#Polyfill)
+before any other code.
+
 ## Compatibility with Express <= 4.x
 
 Path-To-RegExp breaks compatibility with Express <= `4.x`:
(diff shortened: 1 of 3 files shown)


blakeembrey:
When a 2.0 is being looked at, I'll revisit this. For now though, this module is still running tests on node 0.10 and is failing. As much as I didn't want the dependency initially and regret it now, I'm not currently looking at breaking backward compatibility. Thanks!

PR #110, opened 2017-06-28
PATH_REGEXP issue with non-word-characters in parameters

mmschettler:
If a path-parameter contains a non-word-character, than the parameter-name is not correct parsed. 

For example: 
A the path ´´http://.../list-resource/{resource-id}´´(open api spec) ´´.../list-resource/:resource-id´´ will evaluate ´´resource´´ as paramater-name ("-" is valid part of an uri).

The new RegExp recognize only the uri-part-separators (/ ? #) as segment-delimiters.


blakeembrey:
You should run tests locally. This breaks most of the test cases available for a couple of reasons:

1. Not supporting other parameters is a conscious decision. This isn't the open API spec and you should use a different library if that's what you want.
2. This doesn't handle parameter regexp groups.

As of now, there's no intention to change how parameters are matched as the current approach is well defined.

PR #111, opened 2017-06-28
PATH_REGEXP issue with "-" in parameter names

mmschettler:
If a path-parameter contains a "-", than the parameter-name is not correct parsed.

For example:
A the path ´´http://.../list-resource/{resource-id}´´(open api spec) ´´.../list-resource/:resource-id´´ will evaluate ´´resource´´ as paramater-name ("-" is valid part of an uri).

A "-" was added to the RegExp.

Tests done.


frenzzy:
Why don't you want to just use camelCase for parameter names?
```js
const path = '/list-resource/:resourceId';
```

mmschettler:
Because the Open API Specs I use are not my own and I am tired of replacing the parameters each time.

blakeembrey:
See reasons I just wrote in https://github.com/pillarjs/path-to-regexp/pull/110#issuecomment-311754410 please.

PR #114, opened 2017-08-22
Major changes

blakeembrey:
* New option! Ability to set `endsWith` to match paths like `/test?query=string` up to the query string
* New option! Set `delimiters` for specific characters to be treated as parameter prefixes (e.g. `/:test`)
* Remove `isarray` dependency
* Explicitly handle trailing delimiters instead of trimming them (e.g. `/test/` is now treated as `/test/` instead of `/test` when matching)
* Remove overloaded `keys` argument that accepted `options`
* Remove `keys` list attached to the `RegExp` output
* Remove asterisk functionality (it's a real pain to properly encode)
* Change `tokensToFunction` (e.g. `compile`) to accept an `encode` function for pretty encoding (e.g. pass your own implementation)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,8 @@
+sudo: false
 language: node_js
 
 node_js:
-  - "0.10"
-  - "0.12"
   - "4.0"
-  - "4.1"
+  - "stable"
 
 after_script: "npm install coveralls@2 && cat ./coverage/lcov.info | coveralls"
--- a/Readme.md
+++ b/Readme.md
@@ -1,6 +1,6 @@
 # Path-to-RegExp
 
-> Turn an Express-style path string such as `/user/:name` into a regular expression.
+> Turn a path string such as `/user/:name` into a regular expression.
 
 [![NPM version][npm-image]][npm-url]
 [![Build status][travis-image]][travis-url]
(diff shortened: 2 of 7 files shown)


oliviertassinari:
> Remove asterisk functionality (it's a real pain to properly encode)

How should we handle this feature now?

blakeembrey:
@oliviertassinari Did you try the README? https://github.com/pillarjs/path-to-regexp#compatibility-with-express--4x. It's been like this for a while actually, I only hacked the asterisk feature into the 1.x releases around 1.2, but realised that making it act like it use used in Express.js is a mistake.

oliviertassinari:
Thanks, I was curious, I couldn't find the CHANGELOG. I agree, it's better this way.

blakeembrey:
There's a history file here: https://github.com/pillarjs/path-to-regexp/blob/master/History.md. I forgot 1.2 was actually so long ago!

oliviertassinari:
By bad, I don't have the reflex to look for `History.md`. My first reflex when upgrading a lib is to look into the [releases](https://github.com/pillarjs/path-to-regexp/releases), then the [CHANGELOG.md](https://github.com/pillarjs/path-to-regexp/blob/master/CHANGELOG.md). I will also look for this file now. I haven't realized it's a convention in the *express* world.

blakeembrey:
No problem. I believe it's mostly convention from the people who created the projects initially. In every other projects I use releases also 😄

Issue #87, opened 2016-07-28
Make trailing slash optional in splat params.


DylanPiercey:
Hey,

I often find myself writing paths like this:

``` js
app.get('/test/*', ...)

// Which becomes

app.get('/test/(.*)', ...)
```

However 100% of the time what I actually want is this:

``` js
app.get('/test(/?.*)', ...)
```

Is it possible to make a splat only section of the path allow for optional trailing slash?

I realize this would be a breaking change, but I can't perceive any crazy issues with it and it would be pretty convenient for myself and perhaps others.

blakeembrey:
You could do `/:path*` to allow multiple segments. You can also use a regexp if you need something more complex - E.g. `/(.+)*`.

blakeembrey:
It's documented in the README, and this is the reason I had disabled the `*` altogether. It was re-enabled because of the concern for people upgrading in the future from Express 4.x to 5.x. The asterisk doesn't make sense to `/*?` because you couldn't do `/*+` or `/**`. The functionality mirrors `path-to-regexp` 0.x used in Express.js so I'm not sure there'd be any reason to change that.

blakeembrey:
I'm going to close this since the behaviour is set and won't be changing. Even if it did change, there's no guarantee that instead most people really want the current functionality and you're just an outlier. I think the current behaviour is expected, the asterisk is not a param group - it's just an asterisk, but I will keep a look out for similar comments.

blakeembrey:
Ok, so revising this from https://github.com/pillarjs/path-to-regexp/issues/92#issuecomment-250931350. If this change were to happen, it'd be kind of defeating the purpose I added it back to begin with (which was legacy Express.js support). Mostly likely, it'd be compatible with how most people use it - except for that small percent that may actually be using it for literally matching _anything_. For example, the path `/foo////bar` won't match on the splat param (it's looking for well formed paths) while `/*` would match it (it just matches everything). I'd argue these are useful in different cases - but it might be useful to document why this happens. What are your thoughts?

DylanPiercey:
Heres my thinking -

Say you have a path `/thing/:items*` - It makes perfect sense to allow for `/thing` since you said 0 or more items. This is the way it currently works as well. I'm not sure why `/*` is so special, IMO it should just be an unnamed group.

[...]

PR #115, opened 2017-08-23
Better asterisk functionality

blakeembrey:
/cc @DylanPiercey from https://github.com/pillarjs/path-to-regexp/issues/87 for review. Proper `*` behaviour that's incompatible with Express.js (in the strictest sense) but compatible with expectations.

### DIFF
--- a/index.js
+++ b/index.js
@@ -19,9 +19,10 @@ var PATH_REGEXP = new RegExp([
   // Match Express-style parameters and un-named parameters with a prefix
   // and optional suffixes. Matches appear as:
   //
-  // "/:test(\\d+)?" => ["/", "test", "\d+", undefined, "?"]
-  // "/route(\\d+)"  => [undefined, undefined, undefined, "\d+", undefined]
-  '(?:\\:(\\w+)(?:\\(((?:\\\\.|[^\\\\()])+)\\))?|\\(((?:\\\\.|[^\\\\()])+)\\))([+*?])?'
+  // "/:?" => [undefined, undefined, undefined, "?"]
+  // "/:test(\\d+)?" => ["test", "\d+", undefined, "?"]
+  // "/route(\\d+)"  => [undefined, undefined, "\d+", undefined]
+  '(?:\\:(\\w*)(?:\\(((?:\\\\.|[^\\\\()])+)\\))?|\\(((?:\\\\.|[^\\\\()])+)\\))([+*?])?'
 ].join('|'), 'g')
 
 /**
--- a/test.ts
+++ b/test.ts
@@ -1696,7 +1696,7 @@ var TESTS: Test[] = [
     ]
   ],
   [
-    '/.+*?=^!:${}[]|',
+    '/.+*?=^!\\:${}[]|',
     null,
     [
       '/.+*?=^!:${}[]|'
@@ -2218,6 +2218,63 @@ var TESTS: Test[] = [
       [{ foo: 'foo', bar: 'bar' }, '$foo$bar'],
     ]
   ],
+
+   /**
+   * Unnamed modifiers.
+   */
+  [
+    '/:*',
+    null,
+    [
+      {
+        name: 0,
+        prefix: '/',
+        delimiter: '/',
+        optional: true,
+        repeat: true,
+        partial: false,
+        pattern: '[^\\/]+?'
+      }
+    ],
+    [
+      ['', ['', undefined]],
+      ['/', ['/', undefined]],
+      ['/foo/bar', ['/foo/bar', 'foo/bar']]
+    ],
+    [
+      [null, ''],
+      [{ '0': '' }, null],
+      [{ '0': 'foobar' }, '/foobar'],
+      [{ '0': ['foo', 'bar'] }, '/foo/bar'],
+      [{ '0': 'foo/bar' }, '/foo%2Fbar']
+    ]
+  ],
+  [
+    '/foo/:*',
+    null,
+    [
+      '/foo',
+      {
+        name: 0,
+        prefix: '/',
+        delimiter: '/',
+        optional: true,
+        repeat: true,
+        partial: false,
+        pattern: '[^\\/]+?'
+      }
+    ],
+    [
+      ['', null],
+      ['/test', null],
+      ['/foo', ['/foo', undefined]],
+      ['/foo/', ['/foo/', undefined]],
+      ['/foo/bar', ['/foo/bar', 'bar']]
+    ],
+    [
+      [{ '0': 'bar' }, '/foo/bar']
+    ]
+  ]
 ]
 
 /**


blakeembrey:
I'm not sure whether this _should_ open the gates to standalone `/?` and `/+` though. I'm also not sure if we shouldn't just promote `/:path*` instead.

DylanPiercey:
😮! This is awesome, very excited to mess around with this. Potentially can delete some of my own code 😛.

Regarding your last message it may be worth supporting them all. Here are some examples of my current thinking regarding `/?` and `/+` use cases:

```js
// Attach middleware at an optional path without naming it.
match('/my-page/?', middleware)
match('/my-page/:section?', renderSection)

// Attach middleware at an optional arbitrarily deep path without naming it.
match('/my-page/+', middleware)
match('/my-page/:section', renderSection)
match('/my-page/:section/:subsection', renderSubSection)
```

This also applies to hostnames

```js
// Attach middleware at an optional hostname without naming it.
match('?.mydomain.com', middleware)
match('www.mydomain.com', renderMainSite)
match('api.mydomain.com', exposeApi)

// Attach middleware at an optional arbitrarily deep hostname without naming it.
match('+.mydomain.com', middleware)
match('analytics.app.mydomain.com', renderAnalytics)
match('admin.mydomain.com', renderSubSection)
```

Hopefully these examples illustrate some cases where this is useful.
However my personal opinion is that, although I think it aligns with my expectations, this should only be supported if it's possible without adding a ton of complexity to path-to-regexp.

Curious what your thoughts are!

blakeembrey:
The complexity of adding it correctly is super low (see files changed, it's around three lines). I do like the `/:?` as it's less ambiguous. My main concern was path matching with something like `/test?` - it could be confused as a query parameter. At first glance it seems like something like `/my-page/?` may be saying optional `/my-page`.

DylanPiercey:
Actually I think having looked at it a few times it is nice that `/:?` removes ambiguity. I think it could make things simpler as well, essentially:

```js
':' // Exactly one
':?' // Zero or one
':*' // Zero or more
':+' // One or more
``` 

Which I think is intuitive enough. Also regarding it helps disambiguate your previous example, which instead you would write `/:?(test)`. What are your thoughts on that? I am just bouncing ideas around with the hopes of achieving the most flexible/sensible implementation.

DylanPiercey:
One other problem I had before that I wanted to bring up was when you have a leading or trailing group.

For example:

```js
'/test/(.*)' // Will not match '/test'
'*.test.com' // Will not match 'test.com'
```

With the above syntax you would simply do:

```js
'/test/:*' // Will not match '/test'
':*.test.com' // Will not match 'test.com'
```

Is it ugly? I'm not sure 😛

blakeembrey:
@DylanPiercey Do we need a new option to handle "trailing" matches instead of prefixes. I just realised that with hostnames you'd prefer to have it do trailing groups instead of prefixed groups. E.g. `:subdomain?.example.com` isn't possible today because it considers prefixes, not suffixes.

blakeembrey:
One thing that would be nice with that syntax is that if "unnamed groups" matched. E.g. instead of being `(.*)` is was `:(.*)`. I'll think on it though. What do you think is the main advantage of the shortened syntax is? Is adding a name really so bad (it can be a form of documentation - e.g. `:subdomain.example.com`)?

DylanPiercey:
@blakeembrey In 1.7.0 it was possible to do `*.test.com` I believe. 

Also that's what i'm talking about, I think unnamed groups could effectively achieve the same goals while also being familiar and not imposing custom implementations for these operators.

Also it's not the worst thing to force a name. However programming being what it is people will probably have a hard time coming up with names and avoid them, also there is the potential for name conflicts if everything has to be named.

Ultimately I would see people writing `:subdomain.host.com` and `/test/:path` so often that they may end up asking for the shorthands anyway.

blakeembrey:
That shouldn't have ever worked, please feel free to double check. The path matching has always only taken into account prefixes, so it would have failed with `test.com` but not with `x.test.com`. I'm proposing a feature so you can flip it and it'll use suffixes (e.g. `compile(':path*.test.com')()` would properly output `test.com` instead of `.test.com`).

DylanPiercey:
I'll have to take another look at that. It may have been because of the way I was wrapping path-to-regexp. Either way this would be a welcome feature as I'm currently using it 😅

blakeembrey:
Updated the PR. Supporting this functionality is a one character change `\w+` to `\w*`. I'll leave it open for a bit to decide.

DylanPiercey:
Sounds good. If you have any other concerns feel free to ping me as I'd be happy to discuss further.

DylanPiercey:
@blakeembrey any revelations? 😛

blakeembrey:
@dougwilson interested in your feedback also - closer to previous `*` as `:*` (e.g. unnamed and without custom group). Not sure personally how I feel about such a simple delimiter meaning so much, so additional feedback is 👍

mightyiam:
:pick:

jayphelps:
Note: some maintainers prefer new tickets, some prefer keeping the discussion in the original(s). Happy to move to a new ticket if preferred.

---

I did a lot of digging and I _think_ this is the last notable conversation around this topic. I wanted to note that for us, we had to back down to the older 0.1.7 version of this lib because of user backlash, who expect us to support some form of the `/*` wildcard behavior without param names or (.*) regex.

Indeed there have been a few cases I've seen where people differed on their expectations around what exactly this means (how greedy, trailing slashes, etc), but AFAICT a vast majority of the times I've seen it used, those details haven't mattered in practice thankfully.

I think why our folks expect it mostly comes down to "that's what other things support." Some have seen this feature in multiple tools. These three are the most common examples I've seen referenced where they learned this behavior and came to expect it from other tools (like ours):

- Express v4 and earlier
- Cloudflare routes and page rules
- react-router

Our tools don't use express, just path-to-regexp, to be clear. 

I'm personally somewhat conflicted as I can see the arguments against it from a purist standpoint, but at the same time, pragmatically speaking, it seems like this ship has sailed and we're all just trying to fight it for the benefit of a minority of cases at the expense of the majority.

Just adding in my experience in trying to upgrade to the latest , which of course, as always, maintainers are free to ignore since you owe us nothing :shipit:

blakeembrey:
It makes sense to me, I believe it should act like `(.*)` if it is to be added back. It's pretty likely it'll be re-introduced due to discussions in https://github.com/WICG/urlpattern, for compatibility with existing URL patterns. However, there's a few other breaking changes that the discussions in `urlpattern` may introduce so I've been waiting to see whether I should do them all in one go or keep it separate. 

One caveat is that it'll never quite act like express.js 4, where `*` expanded in some odd spots. It'll be pretty close though. Are you aware of any other issues hindering upgrading? The list of other breaking changes for `URLPattern` compatibility are:

1. Case sensitivity (currently insensitive by default)
2. Strict trailing slash (currently non-strict by default)
3. Magic prefixes (i.e. `/:foo?` automatically making the `/` optional)

jayphelps:
@blakeembrey I hadn't seen URLPattern prior. Very interesting indeed!

> Are you aware of any other issues hindering upgrading

Nothing else I had yet found, though it wasn't live for very long so there may have been others.

HansBrende:
@blakeembrey 
> I believe it should act like (.*) if it is to be added back.

Agree! The `*` operator should match any path (whereas `:name*`, and by extension the hypothetical `:*`, appear to only match paths that have _no_ empty path segments in them).

Ideally, need a way to match `/base/ANYTHING` (e.g., `/base/foo`, `/base/foo/`, `/base/foo//bar`, or simply `/base`). Currently the only pleasant way to accomplish this seems to be the following `URLPattern` syntax: `/base/*?`

HansBrende:
Looks like there is an updated PR for this! https://github.com/pillarjs/path-to-regexp/pull/269

HansBrende:
Related discussion in URLPattern: [Named vs. Anonymous wildcard groups have inconsistent and confusing behavior](https://github.com/WICG/urlpattern/issues/163)

HansBrende:
I've created my own PR because the behavior the existing PR will introduce is not pretty. Food for thought: https://github.com/pillarjs/path-to-regexp/pull/270

PR #129, opened 2017-12-06
Clarify named parameters and custom matching parameters in README

mickeyreiss-visor:
* Clarify phrasing
* Dedupe and simplify example

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -61,15 +61,9 @@ re.exec('/test/route')
 //=> ['/test/route', 'test', 'route']
 ```
 
-**Please note:** Named parameters must be made up of "word characters" (`[A-Za-z0-9_]`).
+**Please note:** Parameter names must be made up of "word characters" (`[A-Za-z0-9_]`).
 
-```js
-var re = pathToRegexp('/(apple-)?icon-:res(\\d+).png')
-// keys = [{ name: 0, prefix: '/', ... }, { name: 'res', prefix: '', ... }]
-
-re.exec('/icon-76.png')
-//=> ['/icon-76.png', undefined, '76']
-```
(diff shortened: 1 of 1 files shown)


Readme.md
```diff
@@ -120,18 +112,18 @@ re.exec('/bar/baz')
...
 
-#### Custom Match Parameters
+#### Custom Matching Parameters
 
-All parameters can be provided a custom regexp, which overrides the default (`[^\/]+`).
+All parameters can be provided a custom regexp, which overrides the default match (`[^\/]+`). For example, you can match digits in the path:
 
 ```js
-var re = pathToRegexp('/:foo(\\d+)')
+var re = pathToRegexp('/icon-:foo(\\d+).png')
```
blakeembrey: This example is a bit different to the one above (no optional unnamed group), but happy to merge it anyway since the change is negligible.

Readme.md
```diff
@@ -120,18 +112,18 @@ re.exec('/bar/baz')
...
 // keys = [{ name: 'foo', ... }]
 
-re.exec('/123')
-//=> ['/123', '123']
+re.exec('/icon-123.png')
+//=> ['/icon-123.png', '123', index: 0, input: '/icon-123.png']
```
blakeembrey: Going to remove the extra two params here since none of the other examples use it.

Readme.md
mickeyreiss-visor: @blakeembrey I defer to you on that call, as you understand the nuances of the library better that I do. My main suggestion is to remove duplication and explicitly call out the most important features.

mickeyreiss-visor:
Thanks, @blakeembrey!

PR #135, opened 2018-02-11
ES6 version

theuselessdeveloper:
Rewrote the code in ES6. Passed all tests 👍 
Shoutout to @AlexandruBurlacu for his contribution.

It's my 1st pull request, please be gentle 😄

### DIFF
--- a/index.js
+++ b/index.js
@@ -13,9 +13,17 @@ const DEFAULT_DELIMITERS = './'
  *
  * @type {RegExp}
  */
-const PATH_REGEXP = new RegExp(
-  '(\\\\.)|(?:\\:(\\w+)(?:\\(((?:\\\\.|[^\\\\()])+)\\))?|\\(((?:\\\\.|[^\\\\()])+)\\))([+*?])?',
-  'g')
+const PATH_REGEXP = new RegExp([
+  // Match escaped characters that would otherwise appear in future matches.
+  // This allows the user to escape special characters that won't transform.
+  '(\\\\.)',
+  // Match Express-style parameters and un-named parameters with a prefix
+  // and optional suffixes. Matches appear as:
+  //
+  // "/:test(\\d+)?" => ["/", "test", "\d+", undefined, "?"]
+  // "/route(\\d+)"  => [undefined, undefined, undefined, "\d+", undefined]
+  '(?:\\:(\\w+)(?:\\(((?:\\\\.|[^\\\\()])+)\\))?|\\(((?:\\\\.|[^\\\\()])+)\\))([+*?])?'
+].join('|'), 'g')
 
 /**
  * Parse a string for the raw tokens.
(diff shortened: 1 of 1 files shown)


wesleytodd:
Hey @NulledGravity.  This is closely related to [this conversation](https://github.com/expressjs/express/issues/2755#issuecomment-365011388).   I am not a involved as a maintainer of this particular module, but since it is pretty core to Express I think that discussion has a strong implication on the status of this PR.

On a side note, it might be better, before making your first contribution to a project, to check if the maintainers are interested in moving in that direction.  It will help you not waste time on changes which might not get merged if the maintainer is looking in a different direction.  Just friendly feedback :)

NulledGravity:
Thanks for your feedback @wesleytodd :)

paulmelnikow:
Node 6 is the lowest version that's in long-term support. Based on the linked Express issue, Express will drop support for Node 5 in its next release. It appears the changes in this PR will work in Node 6, and running tests in Node 6 would determine that for sure.

theuselessdeveloper:
That's great news. Thanks @paulmelnikow for updating me. I'll give the changes another look, make sure that it passes all the tests and will update the PR.

I hope we can get this approved.

paulmelnikow:
It would be helpful to know from the maintainers whether they are okay dropping support for Node 4 at this time.

blakeembrey:
Honestly, as is, it can not be accepted. It's deleted all comments, refactors functions into `=>` format and made other changes that appear to be just "because" instead of considering readability. The only ES6 features this package could really use, but not need, is `const`, `let` and string interpolation. It can not delete things we really should keep like the comments or function declarations. Also, we should avoid `.forEach()` callback functions and use `for..of` if we're doing this.

blakeembrey:
In response to https://github.com/pillarjs/path-to-regexp/pull/135#issuecomment-436452087, we also be aware that this module is not _just_ for Express. It's probably more use nowadays in front-end routing frameworks which may want to keep older compatibility than node.js 6, not to mention many other node.js frameworks. We should definitely consider updating our lowest support node.js version as node.js approaches EOL, but as-is there's no major features here that are needed for `path-to-regexp` as it's a fairly simple module - no async, only string manipulation and arrays really.

theuselessdeveloper:
> Honestly, as is, it can not be accepted. It's deleted all comments, refactors functions into `=>` format and made other changes that appear to be just "because" instead of considering readability. The only ES6 features this package could really use, but not need, is `const`, `let` and string interpolation. It can not delete things we really should keep like the comments or function declarations. Also, we should avoid `.forEach()` callback functions and use `for..of` if we're doing this.

I completely agree with you on that. In fact, when a couple of days ago I saw some activity in the thread I have started to refactor my code. I added the comments and replaced the `.forEach()`.

There are still some things that I'm not sure about, but I will push the code hopefully this weekend or the next week and I'd love to hear what you think about it.

blakeembrey:
Added support with https://github.com/pillarjs/path-to-regexp/commit/4124c3ce63519582fb0bafb52763058bc286dcd7.

PR #147, opened 2018-04-20
more precise type check

pflhm2005:
I assume that Number is valid, add a typeCheck function to handle type check.
All type except Regexp|Array|String|Number will throw a invalid type error.
Looking forward to your respond,thank you!

### DIFF
--- a/index.js
+++ b/index.js
@@ -200,12 +200,12 @@ function tokensToFunction (tokens) {
 
 /**
  * Get the type of path
- * 
+ *
  * @param {all} path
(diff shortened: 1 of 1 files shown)


blakeembrey:
@pflhm2005 I'm not going to accept this at this time. `Object.prototype.toString` approaches are generally much slower than `Array.isArray` or `typeof` checks. I also don't feel like the stricter approach is necessary - JavaScript is generally duck typed, and when you want a proper type system it's supported by TypeScript.

PR #176, opened 2018-12-10
Remove `partial` token support, always delimiter

blakeembrey:
Closes https://github.com/pillarjs/path-to-regexp/pull/173.
Closes https://github.com/pillarjs/path-to-regexp/issues/168.

### DIFF
--- a/index.js
+++ b/index.js
@@ -299,10 +299,6 @@ function tokensToRegExp (tokens, keys, options) {
   var delimiter = options.delimiter || DEFAULT_DELIMITER
   var endsWith = [].concat(options.endsWith || []).map(escapeString).concat('$').join('|')
   var route = start ? '^' : ''
-  var endToken = tokens[tokens.length - 1]
-  var isEndDelimited = typeof endToken === 'string'
-    ? endToken[endToken.length - 1] === delimiter
-    : endToken === undefined
 
   // Iterate over the tokens and create our regexp string.
   for (var i = 0; i < tokens.length; i++) {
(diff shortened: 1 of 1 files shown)


test.ts
```diff
@@ -843,17 +829,17 @@ var TESTS: Test[] = [
...
         repeat: false,
-        partial: true,
         pattern: '[^\\/]+?'
       },
       '-bar'
     ],
     [
-      ['/-bar', ['/-bar', undefined]],
+      ['-bar', ['-bar', undefined]],
+      ['/-bar', null],
```
paulmelnikow: This change seems contrary to what I'd expect, which is that the delimiter at the beginning of a path would be required. I think perhaps this is an example where escaping the `/` would accomplish that?

test.ts
```diff
@@ -2037,26 +1976,25 @@ var TESTS: Test[] = [
    * Unnamed group prefix.
    */
   [
-    '/(apple-)?icon-:res(\\d+).png',
+    '\\/(apple-)?icon-:res(\\d+).png',
```
paulmelnikow: It's surprising that this would require a double backslash instead of `'\/(apple-)...'`.

For example, `new RegExp('\/')` and `new RegExp('\\/')` are equivalent.

test.ts
blakeembrey: Yes, to have this be required you would escape the slash making it a regular character instead of a prefix.

test.ts
blakeembrey: Gotcha. This is just because of JavaScript strings. You need to escape the backslash because JavaScript interprets it as an escape character before the library does.

blakeembrey: Note that using a RegExp special character will demonstrate the difference (just slash doesn’t mean anything in RegExps).

RedSparr0w:
:+1:
This seems like a good way to handle this.

paulmelnikow:
Hey! I finally had a chance to integrate this at badges/shields/pull/2644. It's looking great!! Would love to have a release 😁 

(Sorry I lost track of this!)

paulmelnikow:
We're deciding whether to merge this with a GitHub link in `package.json` or wait for the release. Do you think you might be able to merge and cut a release (even a beta)?

blakeembrey:
@paulmelnikow Looking at doing a release this week (probably the weekend). Want to make sure I add a small description and caveats to the README. I'll merge this PR at that point so you'll see it, does that work for you?

paulmelnikow:
Thank you! We decided to merge with the got commit though I am eager to switch back (so our self hosting users avoid complications).

paulmelnikow:
Our tests are passing in 3.0.0!

PR #178, opened 2019-02-26
Add `noValidate` option in `PathFunctionOptions`

helloyou2012:
I want to use this library to generate path template, like this:

```
const ptr = require('path-to-regexp');

const url = '/test/:foo(\\d+)/(.*)';
const keys = [];

const tokens = ptr.parse(url);
const regexp = ptr.tokensToRegExp(tokens, keys);
const toPath = ptr.tokensToFunction(tokens);

const data = {};
keys.forEach(({ name }) => {
  data[name] = name;
});

const path = toPath(data, {
  encode: (v) => `{${v}}`,
  noValidate: true
});

console.log(path); // => /test/{foo}/{0}
```

But it throw match error. So I add `noValidate` option.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -73,6 +73,7 @@ declare namespace pathToRegexp {
      * Function for encoding input strings for output.
      */
     encode?: (value: string, token: Key) => string;
+    skipMatch?: boolean;
   }
 
   export type Token = string | Key;
--- a/index.d.ts
+++ b/index.d.ts
@@ -73,7 +73,10 @@ declare namespace pathToRegexp {
      * Function for encoding input strings for output.
      */
     encode?: (value: string, token: Key) => string;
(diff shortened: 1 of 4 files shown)


index.d.ts
```diff
@@ -73,6 +73,7 @@ declare namespace pathToRegexp {
      * Function for encoding input strings for output.
      */
     encode?: (value: string, token: Key) => string;
+    skipMatch?: boolean;
```
blakeembrey: Would you mind adding some documentation here and mirroring it to the README?

blakeembrey: Maybe calling it `noValidate` or something like that? Do you feel like that's any clearer about intent?

helloyou2012: I rename `skipMatch` to `noValidate` and add the documentation.

PR #187, opened 2019-04-21
 Fix case sensitivity when generating paths

boringame:


blakeembrey:
This should probably be an option with the same defaults as the normal `ToRegexp` method.

PR #188, opened 2019-04-23
tokensToFunction sensitive flags(options)

boringame:


boringame:
Anybody home?

blakeembrey:
Do you want to undo all the unrelated changes made?

boringame:
Those Spaces could be an editor plug-in problem.Could you please check if there are any other questions? If not, I will resubmit them.

blakeembrey:
@boringame That's my only concern.

PR #191, opened 2019-07-10
TokensToFunctionOptions

boringame:


pimlie:
Would it maybe be possible to backport this fix to the v1 branch which is used by vue-router? Or is v1 not supported anymore at all?

Currently experiencing [this issue](https://github.com/vuejs/vue-router/issues/2925) but due to breaking changes in path-to-regexp v2+ vue-router is stuck at v1.x until its next major release (which might be a while).

PR #197, opened 2019-09-25
Use ES2015 module syntax to enable tree shaking

frenzzy:
This PR allow users to use `path-to-regexp` with all the benefits of ES Modules.

**How to use:**

```js
import * as pathToRegexp from 'path-to-regexp'

// pathToRegexp.pathToRegexp(path, keys?, options?)
// pathToRegexp.parse(path)
// pathToRegexp.compile(path)
```
or
```js
import { pathToRegexp, parse, compile } from 'path-to-regexp'

// pathToRegexp(path, keys?, options?)
// parse(path)
// compile(path)
```

or

```js

// pathToRegexp.pathToRegexp(path, keys?, options?)
// pathToRegexp.parse(path)
// pathToRegexp.compile(path)
```
or
```js
const { pathToRegexp, parse, compile } = require('path-to-regexp')

// pathToRegexp(path, keys?, options?)
// parse(path)
// compile(path)
```

**How to migrate from v3 to v4:**

```diff
  const pathToRegexp = require('path-to-regexp')

- // pathToRegexp(path, keys?, options?)
+ // pathToRegexp.pathToRegexp(path, keys?, options?)
  // pathToRegexp.parse(path)
  // pathToRegexp.compile(path)
```

**Why:**

It allows to make our client-side bundles smaller.
For example see how ESM version affects [`universal-router`](https://github.com/kriasoft/universal-router) package:

| File Size (bytes) | Before | After | Saved |
|-|-|-|-|
| universal-router.min.js.gz | 2490 | 2069 | 17% |
| universal-router.min.js | 5598 | 4436 | 21% |

---

Also it allows to use the package in modern browsers without any transformations (no compilation or build step required) :tada:
```js
<script type="module">
  import * as pathToRegexp from 'https://unpkg.com/path-to-regexp?module'

  // pathToRegexp.pathToRegexp(path, keys?, options?)
  // pathToRegexp.parse(path)
  // pathToRegexp.compile(path)
</script>
```

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,94 +1,90 @@
-declare function pathToRegexp (path: pathToRegexp.Path, keys?: pathToRegexp.Key[], options?: pathToRegexp.RegExpOptions & pathToRegexp.ParseOptions): RegExp;
+export function pathToRegexp (path: Path, keys?: Key[], options?: RegExpOptions & ParseOptions): RegExp;
 
-declare namespace pathToRegexp {
-  export interface RegExpOptions {
-    /**
-     * When `true` the regexp will be case sensitive. (default: `false`)
-     */
-    sensitive?: boolean;
-    /**
-     * When `true` the regexp allows an optional trailing delimiter to match. (default: `false`)
-     */
-    strict?: boolean;
-    /**
-     * When `true` the regexp will match to the end of the string. (default: `true`)
-     */
-    end?: boolean;
-    /**
-     * When `true` the regexp will match from the beginning of the string. (default: `true`)
-     */
-    start?: boolean;
-    /**
-     * Sets the final character for non-ending optimistic matches. (default: `/`)
-     */
-    delimiter?: string;
-    /**
-     * List of characters that can also be "end" characters.
-     */
-    endsWith?: string | string[];
-    /**
-     * List of characters to consider delimiters when parsing. (default: `undefined`, any character)
-     */
-    whitelist?: string | string[];
-  }
-
-  export interface ParseOptions {
-    /**
-     * Set the default delimiter for repeat parameters. (default: `'/'`)
-     */
-    delimiter?: string;
-  }
-
-  export interface TokensToFunctionOptions {
-    /**
-     * When `true` the regexp will be case sensitive. (default: `false`)
-     */
-    sensitive?: boolean;
-  }
-
+export interface RegExpOptions {
   /**
-   * Parse an Express-style path into an array of tokens.
+   * When `true` the regexp will be case sensitive. (default: `false`)
    */
-  export function parse (path: string, options?: ParseOptions): Token[];
-
+  sensitive?: boolean;
   /**
-   * Transforming an Express-style path into a valid path.
+   * When `true` the regexp allows an optional trailing delimiter to match. (default: `false`)
    */
-  export function compile <P extends object = object> (path: string, options?: ParseOptions & TokensToFunctionOptions): PathFunction<P>;
-
+  strict?: boolean;
   /**
-   * Transform an array of tokens into a path generator function.
+   * When `true` the regexp will match to the end of the string. (default: `true`)
    */
-  export function tokensToFunction <P extends object = object> (tokens: Token[], options?: TokensToFunctionOptions): PathFunction<P>;
-
+  end?: boolean;
   /**
-   * Transform an array of tokens into a matching regular expression.
+   * When `true` the regexp will match from the beginning of the string. (default: `true`)
    */
-  export function tokensToRegExp (tokens: Token[], keys?: Key[], options?: RegExpOptions): RegExp;
-
-  export interface Key {
-    name: string | number;
-    prefix: string;
-    delimiter: string;
-    optional: boolean;
-    repeat: boolean;
-    pattern: string;
-  }
-
-  interface PathFunctionOptions {
-    /**
-     * Function for encoding input strings for output.
-     */
-    encode?: (value: string, token: Key) => string;
-    /**
-     * When `false` the function can produce an invalid (unmatched) path. (default: `true`)
-     */
-    validate?: boolean;
-  }
-
-  export type Token = string | Key;
-  export type Path = string | RegExp | Array<string | RegExp>;
-  export type PathFunction <P extends object = object> = (data?: P, options?: PathFunctionOptions) => string;
+  start?: boolean;
+  /**
+   * Sets the final character for non-ending optimistic matches. (default: `/`)
+   */
+  delimiter?: string;
+  /**
+   * List of characters that can also be "end" characters.
+   */
+  endsWith?: string | string[];
+  /**
+   * List of characters to consider delimiters when parsing. (default: `undefined`, any character)
+   */
+  whitelist?: string | string[];
 }
 
-export = pathToRegexp;
+export interface ParseOptions {
+  /**
+   * Set the default delimiter for repeat parameters. (default: `'/'`)
+   */
+  delimiter?: string;
+}
+
+export interface TokensToFunctionOptions {
+  /**
+   * When `true` the regexp will be case sensitive. (default: `false`)
+   */
+  sensitive?: boolean;
+}
+
+/**
+ * Parse an Express-style path into an array of tokens.
+ */
+export function parse (path: string, options?: ParseOptions): Token[];
+
+/**
+ * Transforming an Express-style path into a valid path.
+ */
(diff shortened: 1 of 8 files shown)


wesleytodd:
I am pretty solidly against this change.  There are a few reasons:

1. This will not work in node with the current ES modules implementation (requires `.mjs` or `"type": "module"`) see: https://nodejs.org/api/esm.html
2. The node ESM implementation is still experimental
3. Introduces a build tool dep (generally best avoided)
4. Means the node version (the current version and most important place this is used) is a compiled file, making debugging and readability much more difficult

Unfortunately I think the only approach to this that IMO is reasonable at the moment is to maintain a fork which exports ESM.  This could be something that this package does, but it would have to have two maintained branches and two packages published.  This would increase the publish complexity greatly.

I understand how much this whole situation sucks, but making everything more complicated for a *very small* short term gain is just not worth it IMO.

frenzzy:
Hi @wesleytodd, thanks for your feedback.

> 1. This will not work in node with the current ES modules implementation (requires .mjs or "type": "module") see: https://nodejs.org/api/esm.html
> 2. The node ESM implementation is still experimental

This PR proposes ES modules mostly for browsers, Node.js will continue use commonjs version.
Also it will be much easier in the future enable ES modules for Node.js too without introducing a braking change.

> 3. Introduces a build tool dep (generally best avoided)

Agree, but it is better than maintain two almost identical versions of the file. Also rollup is a very small dev dependency. Published `path-to-regexp` package version is continue to be completely dependencies free.

> 4. Means the node version (the current version and most important place this is used) is a compiled file, making debugging and readability much more difficult

Disagree, it absolutely the same file with changed exports from `export { x }` to `exports.x = x`. All other code is untouched, comments formatting etc are preserved so it does not affect debugging and readability.

---

Overall this PR is focused to provide ESM version in addition to currently provided commonjs version and in my opinion does not introduce any disadvantages for existed Node.js users. Instead it adds a useful feature for client-side libraries which use `path-to-regexp` as a dependency.

---

I agree that one of alternative options is to provide ESM and CJS versions separately in two different packages, but I would like to hear @blakeembrey opinion on this point. The best option is to have a single maintainer for both module systems, so it should be up to owner to decide.

IMO it is much simpler to publish the package with one more `.js` file included instead of publish two packages.

index.d.ts
```diff
@@ -1,94 +1,90 @@
...
+  /**
+   * When `false` the function can produce an invalid (unmatched) path. (default: `true`)
+   */
+  validate?: boolean;
 }
 
-export = pathToRegexp;
+export type Token = string | Key;
+export type Path = string | RegExp | Array<string | RegExp>;
+export type PathFunction <P extends object = object> = (data?: P, options?: PathFunctionOptions) => string;
```
frenzzy: The actual change in this file is:
```diff
- declare function pathToRegexp ...
+ export function pathToRegexp ...

- declare namespace pathToRegexp {
    export ...
    ...
    export ...
- }
```
and
```diff
- interface PathFunctionOptions ...
+ export interface PathFunctionOptions ...
```
all other formatting is preserved.

frenzzy:
Oh, I completely forgot about isomorphic code when CJS version is used on server-side and ESM version on client-side. It is highly desirable to provide both versions from a single package to simplify creation of universal applications.

wesleytodd:
I have had this discussion too many times for me to justify spending more time writing detailed responses, but I stand by my original opposition.  As @blakeembrey is the decision maker here I will leave the conversation in his capable hands.

frenzzy:
Hi @blakeembrey, could you please share your opinion on this?

blakeembrey:
@frenzzy I'm generally OK with this, but I'd like to see what shakes out of the NodeJS ESM implementation before I release a couple of back-to-back major releases. If I were going to do this, I'd probably look at refactoring the source code into TypeScript and compiling it to two module systems instead (e.g. https://github.com/blakeembrey/free-style/blob/7111fa02e65921b5836da2a838438464d8bbc900/package.json#L16). If this were a larger project (in terms of SLOC) I'd probably be more inclined to go ahead. Have you done any tests on how many bytes you'd reasonably be saving here?

frenzzy:
@blakeembrey thanks for your response.

Yes, I measured how may bytes will be saved if I need only `pathToRegexp` export from `path-to-regexp` package, in this case `compile` and `tokensToFunction` will be tree shaked away. I attached this info to the PR description using the example of [Universal Router](https://github.com/kriasoft/universal-router):

| File Size (bytes) | Before | After | Saved |
|-|-|-|-|
| universal-router.min.js (uglify) | 5598 | 4436 | 21% |
| universal-router.min.js.gz (uglify+gzip) | 2490 | 2069 | 17% |

In case if you are asking how `pathToRegexp` size changed with tree shaked `compile` and `tokensToFunction` exports:

| File Size (bytes) | Before | After | Saved |
|-|-|-|-|
| path-to-regexp/index.js | 9504 | 6976 | 27% |
| path-to-regexp/index.min.js (uglify) | 3540 | 2320 | 34% |
| path-to-regexp/index.min.js.gz (uglify+gzip) | 1509 | 1092 | 28% |

I am not quite understand why we have to wait until ESM in Node.js become stable, it is just matter of their internal implementation. Module syntax is a standard in JavaScript since June 2015 ([ECMAScript 2015, 6th Edition](https://en.wikipedia.org/wiki/ECMAScript#ES2015)). All modern browsers supports ESM, syntax will not change. When Node.js will officially support ESM we don't need to introduce any new breaking changes.

Also I saw your responses in other issues where you said that `path-to-regexp` widely used in different client-side routing solutions. Don't you think it's worth releasing this featrue without waiting for Node.js? :)

About migration the source code to TypeScript, I understand all the benefits and consequences. In my opinion we (open source community) should keep the entry threshold for new developers as low as possible. Do you remember a time when you was a beginner? They can read and understand the code, send a PR or file an issue without the need to spend extra time learning TypeScript. `path-to-regexp` package already have tests in TS and defenetions. Isn't that quite enough? Of course, this is a very subjective point of view, and it is up to you to decide.

My question is do you really want to release this feature only together with rewriting the source code into TypeScript? If so, I will just close this PR.

wesleytodd:
I would like to throw my support behind *not* re-writing this in TS.  As a small library there is almost no benefit to having it in TS, and there are many downsides.  Again, I am not the one actively maintaining it, so maybe there productivity benefits for @blakeembrey specifically.  If those outweigh the complexity of adding a build step to a module which did not previously have one, then maybe it makes sense.

As for the points made about not waiting for node, this is something I have a stake in.  As Express uses this module, if we don't have full support and a compatible upgrade path it would most likely mean we cannot adopt that change.  So for that reason I am very strongly behind waiting until the path is clear for node support.

blakeembrey:
> I would like to throw my support behind not re-writing this in TS.

There's a large implication that using TypeScript is some sort of rewrite. It's still JavaScript, no different than if you'd used Babel or something else. Even this build system this PR is introducing is overhead here.

> As a small library there is almost no benefit to having it in TS, and there are many downsides.

I'd love to hear the downsides enumerated, I'm not very familiar with any. Adding types to it makes it much easier to pop back into the project without context, make sure the changes I make aren't breaking changes and reduce the amount of development friction (e.g. thinking about minimum ES targets, reduced tooling to maintain `.d.ts` and `.js` in sync, etc). Conceptually this PR is a build step just as TypeScript would be, and as long as we intend to support two module systems there will be some kind of build step. 

> In my opinion we (open source community) should keep the entry threshold for new developers as low as possible.

This argument feels disingenuous. Maybe I underestimate the overhead of types, but this argument assumes that the thing stopping a contribution is seeing `.ts` somehow. I'm always a beginner in something, if I wasn't I'd stay writing jQuery and ES3. Maybe it's worth seeing the difference in source code to judge for yourself for jumping to assumptions, it's possible a beginner may be even more willing to learn something new than stick to old ways. I can't judge for everyone, but I've brought a number of people up to speed on TypeScript at my current company and we've had great success - people don't write `.js` anymore and TypeScript safety increases development productivity.

blakeembrey:
Here's an example of the project changes to TypeScript: https://github.com/pillarjs/path-to-regexp/pull/202. The only major additions is `interface` and type parameters to arguments. Just this migration caught three bugs - two in the type definition (non-nullable types, incorrect interface) and one in code (using `null` instead of strings, could be argued it's also a type bug). The benefit of keeping this now would be that I can write modern code (e.g. `const` and `let`) and know it won't break any ES5 users (I know I've previously written `const` out of habit before, but thankfully old node.js tests catch that).

blakeembrey:
Added with https://github.com/pillarjs/path-to-regexp/commit/4124c3ce63519582fb0bafb52763058bc286dcd7.

PR #201, opened 2019-10-24
Changed coverage tool to nyc

webdevium:


.travis.yml
```diff
@@ -5,4 +5,4 @@ node_js:
   - "6"
   - "stable"
 
-after_script: "npm install coveralls@2 && cat ./coverage/lcov.info | coveralls"
+after_script: "npm install coveralls && npm run coverage | coveralls"
```
blakeembrey: Can we keep `coveralls` pinned to the major version, there's no guarantee the API won't change later.

webdevium: No problem. Fixed!

PR #202, opened 2019-10-26
Add TypeScript to package

blakeembrey:


frenzzy:
What this section in package.json does? Shouldn't it be updated as removed?
https://github.com/pillarjs/path-to-regexp/blob/4124c3ce63519582fb0bafb52763058bc286dcd7/package.json#L28-L32
...to use `index.ts` or `dist/index.js`

Also why you removed `index.ts` from `files` section of package.json? Source maps directs to this file, would be nice to publish it as well, don't you think?

blakeembrey:
There is still more work to do, which is why it hasn’t been published. Please be patient, my day job is not working for you.

blakeembrey:
@frenzzy Released with https://github.com/pillarjs/path-to-regexp/releases/tag/v4.0.0, I got a few more things into the release that were breaking changes (specifically balanced brackets for nested non-capturing groups and `normalizePathname` which removes the automatic `decode` behavior).

blakeembrey:
> Source maps directs to this file, would be nice to publish it as well, don't you think?

No, it shouldn't be necessary. Source maps should be containing the source since it's set to `inlineSources`. I haven't had that be an issue since enabling that in other packages.

PR #203, opened 2019-10-27
Add TypeScript and ES2015 modules

frenzzy:
Continues #202 (closes #202, closes #197, closes #190)

```diff
  const pathToRegexp = require('path-to-regexp')

- // pathToRegexp(path, keys?, options?)
+ // pathToRegexp.pathToRegexp(path, keys?, options?)
  // pathToRegexp.parse(path)
  // pathToRegexp.compile(path)
```


blakeembrey:
Thanks @frenzzy! Let me give this a little time and see if we can get @wesleytodd more on board, ultimately it does increase my ability to maintain the package but it is a departure from the standards of this organization.

wesleytodd:
Well....😂....I have stated my opinions on this more than once.  If @blakeembrey feels that this makes it more maintainable I will not stand in the way.  But I don't have to agree.

As for "standards of this organization", I think it is more about if we have enough maintainers to help out.  In this case, you are the only maintainer here who actively writes TS. That is a risk to me.  In the current state, if you were to leave the project or worse (bus factor...), I could step in and continue to move the project forward.  If it was written in TS I could not be an effective maintainer.

This whole thing is why I have been working with the Package Maintenance WG and the OpenJSF to help make sure Express and it's ecosystem have more support.  If we can get more contributors again, and have a healthy growth story, then I think it would be great to have TS (or flow or whatever other language people want to write and compile to JS).  But I am not interested in learning TS just so I can offer more of my free time to maintain these projects.

blakeembrey:
> if you were to leave the project or worse (bus factor...)

It could always be reverted back to JavaScript. TypeScript ultimately emits `.js` and `.d.ts`, which is just like what we had previously written by hand before. Definitely understand this stance, and appreciate you won't index too much on it.

> But I am not interested in learning TS just so I can offer more of my free time to maintain these projects.

I'd actually love to connect on this, maybe offline? I've talked with a few people now (some who have since converted to using TypeScript full-time in OSS) whose feelings are similar, that TypeScript will take more time. Is it just the feeling of having to learn types (a one-off cost which would not mean offering _more_ free time overall) or do you feel like maintaining TypeScript is more expensive than JavaScript?

wesleytodd:
> It could always be reverted back to JavaScript.

Sure, but in this hypothetical, I would then have to revert it back, taking time from other more productive things I could be working on.  It would also then be a wasted effort on both sides.  A much better scenario would be to find a new contributor who has already invested in TS.

> I'd actually love to connect on this, maybe offline?

Sure, I am not sure if this fits in the Express TC meeting scope, but we could add the topic to the next agenda.  If not there I would be happy to do a hangout or zoom or whatever.

> Is it just the feeling of having to learn types

I have programmed types before.  It is way more than just types, it's a new toolchain, new paradigms, etc.

> not mean offering more free time overall

This is the subjective part.  Every language I have programmed in which were typed have been less productive for me than JavaScript.  Most new tools chains I have used has cost time and been marginally more productive than the old one.  Just like devices, staying at the bleeding edge has huge cost, and I prefer to be more conservative in when I update.

So the point is, even if it is marginally more productive it would take years to fully recoup the cost, and that is a risk I am not comfortable taking at this time.

blakeembrey:
Merged with https://github.com/pillarjs/path-to-regexp/commit/4124c3ce63519582fb0bafb52763058bc286dcd7.

PR #208, opened 2019-11-15
Add documentation for modified param with custom regexp

Khartir:
It is not quite clear where the modifier goes, when modifying a parameter with a custom regexp. Took me a while to figure it out. This should clear this up.

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -152,6 +152,21 @@ regexpWord.exec("/users");
 
 **Tip:** Backslashes need to be escaped with another backslash in JavaScript strings.
 
+##### Modified Custom Matching Parameters
+
+When using a modifier whith a custom regexp, the modifier is added after the regexp:
+
+```js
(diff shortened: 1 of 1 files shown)


blakeembrey:
@Khartir I'd prefer to do two things if it makes sense to you:

1. Make using modifiers in the incorrect place an error
2. Update an existing example to be optional

Would this have helped?

Khartir:
Sounds great.

blakeembrey:
Do you want to try with 6.0 (errors added) and review the latest README? Let me know if it clarifies things.

Khartir:
I can't find an example (or explanation) of the correct order of modifiers and custom matching parameters in the README. Am I missing something or did that not make it in?

The error lokks good.

blakeembrey:
How is https://github.com/pillarjs/path-to-regexp#modifiers?

Khartir:
Great :+1:

Issue #219, opened 2020-02-28
What caused the size increase?


StephenTangCook:
We haven't updated our npm dependency since v3.0.0 and now at v6.1.0 there is a nontrivial size increase. npm.js reports the unpacked size of 3.2.0 to be [30.9kB](https://www.npmjs.com/package/path-to-regexp/v/3.2.0) and 4.0.0 to be [470kB](https://www.npmjs.com/package/path-to-regexp/v/4.0.0).
Curious what caused the jump?

jonchurch:
It looks to be the way things are compiled w/ Typescript now?

Running `npm pack --dry-run` on v6.1.0 shows the tarball unpacked as 476.3 kB.

v.s. on v3.2.0 it's 30.9kb

Here's a gist w/ the output for each version's `npm pack --dry-run`:
https://gist.github.com/jonchurch/c22c4f87f25377d800967772a55ba4db

I don't honestly know much about typescript, but it appears that the file `src/index.spec.ts` is also included in the build. I assume that is a testing only file, and if that's the case, then it should not be bundled.

Ignoring that file cuts the bundled size down to 109.8 kB unzipped.

blakeembrey:
The NPM pack size will always be a bit misleading here since it counts everything. For example, we publish both ES2015 and CommonJS which also doubles the size. However, the package has grown in size each version it's doing a bit more each time by exposing simpler ways to use the package (e.g. `pathToRegexp` -> `match`).

jonchurch:
Closing as answered, and the pack size has been reduced w/ https://github.com/pillarjs/path-to-regexp/pull/220

PR #220, opened 2020-02-29
Ignore .spec.ts files when compiling

jonchurch:
Relates to #219 

Assuming the `.spec.ts` file are not intended to be part of the distribution, they should be ignored when compiling.

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -13,5 +13,6 @@
     "sourceMap": true,
     "inlineSources": true
   },
-  "include": ["src/**/*"]
+  "include": ["src/**/*"],
+  "exclude": ["src/**/*.spec.ts"]
 }


jonchurch:
To be clear, it looks like the problem is that we are publishing the spec file.

This can also be solved in a different way, where test files are in a specific `test` folder, which is not included in the `files` list that is published to npm.

blakeembrey:
Dumb question, but can we use the NPM config to ignore these files? Last I recall you couldn't combine `.npmignore` with `files`, or put negations in `files`, but those would be a tiny bit nicer.

jonchurch:
A `.gitignore` entry would do the trick, npm respects that when publishing.

jonchurch:
An `.npmignore` does work, but then it [overrides anything in `.gitignore`](https://docs.npmjs.com/misc/developers?_ga=2.28078709.817769899.1582996859-1353226532.1582996859#keeping-files-out-of-your-package) when it comes to publishing.

Since the package is whitelisting with the `files` field of `package.json`, it's less of a concern, but still something to keep in mind.

blakeembrey:
@jonchurch Thanks for exploring! Sorry to put you through that. Let's merge and we can release as is.

jonchurch:
Thanks to @StephenTangCook for asking the question!

blakeembrey:
We may need to revert this or do additional work. I realized the other week that by excluding these files, it’s not actually using the right settings for type checking. Not a problem right now, but we should probably have a specific build config instead of changing this one.

PR #225, opened 2020-06-28
Add named capturing groups to regexpToRegexp method

DamianKu:
I was looking into fixing issues raised on express repository [#4277](https://github.com/expressjs/express/issues/4277).

I modified regular expression in `regexpToRegexp` to match not only `(` but also `(?<*>` <sup>probably regexp could be better :)</sup>.

No change with normal regexp matching, `router.get(/\/regex\/(.+)/)` -> `localhost/regex/testData` will result in `{"0":"testData"}`.

With named capturing groups, `router.get(/\/regex2\/(?<groupname>.+)/)` -> `localhost/regex2/testData` will result in `{"groupname":"testData"}`.
Mixed matched with named capturing groups, `router.get(/\/(.+)\/(?<groupname>.+)\/(.+)/)` -> `localhost/testRegex/testData/extraStuff"` will result in `{"0":"testRegex","2":"extraStuff","groupname":"testData"}`.

All tests are passing but I didn't add new tests to cover this functionality yet.
Would be great to get some feedback from you guys if I'm even looking in good place or is it worth working on it.


blakeembrey:
This looks really great to me, thanks! From my understanding this will work in all current projects because the named group doesn't remove it from the array of RegExp results (e.g. this code depends on the position of keys in the result: https://github.com/pillarjs/path-to-regexp/blob/98ab888a26762c1ba10e93a4679b9773e050d1ed/src/index.ts#L405). If you want to add a couple of tests around it we should be good to merge this.

For getting this feature into the current version of Express, you'll probably need to make this PR against an older branch here: https://github.com/pillarjs/path-to-regexp/blob/0.1.x/index.js

However, I believe this it can land in the next Express.js release with the current implementation.

src/index.ts
```diff
@@ -453,13 +453,17 @@ export type Token = string | Key;
 function regexpToRegexp(path: RegExp, keys?: Key[]): RegExp {
   if (!keys) return path;
 
-  // Use a negative lookahead to match only capturing groups.
-  const groups = path.source.match(/\((?!\?)/g);
+  const groups = path.source.match(/\((\?<.*?>)?(?!\?)/g);
```
blakeembrey, line 456:
```suggestion
  const groups = path.source.match(/\((?:\?<(.*?)>)?(?!\?)/g);
```

Something like this will resolve the comment you have below.

DamianKu: Yeah, non-capturing groups, I tried that. Had no luck so I'm guessing, I made some mistake :) 

Will try this :)

DamianKu: > Something like this will resolve the comment you have below.

That didn't work 
```
$ curl http://localhost:1337/testRegex/testData/extraStuff -w "\n"
{"0":"testRegex","2":"extraStuff","(?<groupname>":"testData"}
```

DamianKu:
Test is failing on node v8 as Named capture groups are behind '--harmony' runtime flag.

image

blakeembrey:
@DamianKu You'll probably need to put the tests behind a version flag by checking the node.js version in the test suite.

src/index.ts
blakeembrey: Oh yeah, that makes sense. It's happening with `.match` which is only getting the matches and not the groups. We probably need to use `regexp.exec` in a loop instead.

There's also https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/matchAll but it's not supported on earlier node.js versions or browsers.

DamianKu: Using `exec` worked :)

Looping through `exec` results doesn't look nice tho.
Any idea how to make it nicer?

DamianKu:
> @DamianKu You'll probably need to put the tests behind a version flag by checking the node.js version in the test suite.

@blakeembrey 
How I would go about doing that?
Do you have any example in other your projects that I could investigate? :)

blakeembrey:
> How I would go about doing that?

You can read the node.js version here: https://nodejs.org/api/process.html#process_process_version. And use something like https://www.npmjs.com/package/semver to compare whether you want to run it. Put it within an `if` condition to define the test.

Here's an example (albeit using TypeScript): https://github.com/TypeStrong/ts-node/blob/866bce6a175ec124fe62f269b22e91c92b40f875/src/index.spec.ts#L137

DamianKu:
Hah OK 😁 I'm aware of that. I don't know why I was thinking that you meant
something more elaborate

DamianKu:
I added the `semver` check after the initialization of `TEST` variable.

If you prefer I can do ternary in `TEST` like so:

```typescript
const TESTS: Test[] = [
    ...,

    ...(gte(process.version, "10.0.0")
    ? [
        [...test...],
        [...test...]
      ]
    : ([] as any)),
    
    ...,
];


```

src/index.ts
```diff
@@ -453,19 +453,21 @@ export type Token = string | Key;
...
+  while (execResult) {
+    keys.push({
+      // Use parenthesized substring match if available, index otherwise
+      name: execResult[1] || index,
+      prefix: "",
+      suffix: "",
+      modifier: "",
+      pattern: ""
+    });
+    index++;
```
blakeembrey, line 469: It would probably be good to stay consistent with the behavior in string tokens. In that method, we're only incrementing the counter each non-named group: https://github.com/pillarjs/path-to-regexp/blob/8b0ae94f9ada4705a6144d639088b058f52bf75d/src/index.ts#L189.

That would mean you probably want `execResult[1] || index++`. We should add a mixed test of named and unnamed groups to make sure it doesn't break.

DamianKu: OK will do that. 

Does it mean that this `/\/(.+)\/(?<groupname>.+)\/(.+)/` called with `localhost/testRegex/testData/extraStuff` should result in `{"0":"testRegex","1":"extraStuff","groupname":"testData"}` ?

blakeembrey:
Thanks @DamianKu! This looks great.

PR #227, opened 2020-07-08
Update docs about strict

Janpot:
If I understand correctly `strict` means it won't allow a trailing slash to match


blakeembrey:
Thanks for the catch. There documentation is also shared in TypeScript here (sorry!): https://github.com/pillarjs/path-to-regexp/blob/98ab888a26762c1ba10e93a4679b9773e050d1ed/src/index.ts#L503.

blakeembrey:
Awesome, thanks for the contribution @Janpot!

PR #256, opened 2021-07-21
Improve documentation

pinksynth:
Hello, thank you for this awesome library!

In the proposed changes, I moved the `pathToRegexp` function documentation to its own section so that it follows the structure of the other functions. I added a little clarification, and an example for using named parameters with `match` (it took me a while to understand how to use this lib for named capture groups).

Hope this is helpful!

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -26,9 +26,13 @@ const { pathToRegexp, match, parse, compile } = require("path-to-regexp");
 // compile(path)
 ```
 
+### Path to regexp
+
+The `pathToRegexp` function will return a regular expression object based on the provided `path` argument. It accepts the following arguments:
+
 - **path** A string, array of strings, or a regular expression.
-- **keys** An array to populate with keys found in the path.
-- **options**
+- **keys** _(optional)_ An array to populate with keys found in the path.
+- **options** _(optional)_
   - **sensitive** When `true` the regexp will be case sensitive. (default: `false`)
   - **strict** When `true` the regexp won't allow an optional trailing delimiter to match. (default: `false`)
   - **end** When `true` the regexp will match to the end of the string. (default: `true`)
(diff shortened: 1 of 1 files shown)


Readme.md
```diff
@@ -194,6 +198,21 @@ fn("/invalid"); //=> false
 fn("/user/caf%C3%A9"); //=> { path: '/user/caf%C3%A9', index: 0, params: { id: 'café' } }
 ```
 
+The `match` function can be used to custom match named parameters (as of Node 10.0.0). For example, this can be used to whitelist a small number of valid paths:
```
jonchurch, line 201: I personally don't know what you mean by "(as of Node 10.0.0)" here. 
I ran the example code under Node v9.11.2 and got the same output. Can you elaborate for me?

jonchurch (changes requested):
I think this looks solid. Can you revert the header changes, unless you have a reason they are included? I'm assuming it was an autofix markdown linter, and not intentional.

Readme.md
pinksynth: Hmm, I must've been confusing that limitation with that of another function:
https://github.com/pillarjs/path-to-regexp/pull/225

Will make the change.

pinksynth:
> Can you revert the header changes, unless you have a reason they are included?

@jonchurch I was treating the "Parameters" section, "Named parameters" section, etc, as parts of the `pathToRegexp` documentation, so I increased their header level. I think those sections apply to other functions as well though. Latest version should be good.

jonchurch (approved):
LGTM. It might take some time before the maintainer merges this in, assuming they don't have requests of their own. But thank you for your time and efforts to improve the documentation!

Readme.md
```diff
@@ -194,6 +198,21 @@ fn("/invalid"); //=> false
 fn("/user/caf%C3%A9"); //=> { path: '/user/caf%C3%A9', index: 0, params: { id: 'café' } }
 ```
 
+The `match` function can be used to custom match named parameters. For example, this can be used to whitelist a small number of valid paths:
```
blakeembrey, line 201:
```suggestion
The `match` function can be used with custom match named parameters. For example, this can be used to whitelist a small number of valid paths:
```

Is the goal of this section just to demonstrate you can use all the same paths as the previous documentation sections? Would it be better just to say that explicitly?

pinksynth: The goal is mostly just to demonstrate how the lib could be used to name some matched parameters. With the existing docs, it took me a while to understand that `:tab(about|photos)` would actually provide a `tab` param with the matched value.

PR #269, opened 2022-01-13
Add * wildcard support.

wanderview:


wanderview:
@blakeembrey PTAL.  This is the first in a series of pull requests to add changes/features necessary for urlpattern.

This first PR adds `*` wildcard support.

Note, I was going to squash the commits and add a better commit message, but it seems some newer version of a tool you use in the commit hook added a ton of commas at the end of arrays throughout the codebase.  Not sure if you have a way to avoid that.  Anyway, happy to try to reformat this however you would like.

Thanks.

src/index.ts
```diff
@@ -190,7 +208,7 @@ export function parse(str: string, options: ParseOptions = {}): Token[] {
         prefix,
         suffix: "",
         pattern: pattern || defaultPattern,
-        modifier: tryConsume("MODIFIER") || ""
+        modifier: tryConsumeModifier() || "",
```
HansBrende, line 211: @wanderview Are you intentionally allowing an asterisk to be followed by another modifier? URLPattern doesn't support that AFAIK...

blakeembrey:
I rebased this PR on master and force pushed to your branch (just a heads up on that one) to fix the formatting issue. Sorry about that!

src/index.spec.ts
```diff
@@ -1119,6 +1119,69 @@ const TESTS: Test[] = [
...
+      {
+        name: 0,
+        prefix: "/",
+        suffix: "",
+        modifier: "",
+        pattern: ".*",
+      },
+    ],
+    [
+      ["/", ["/", ""]],
```
blakeembrey, line 1138: Adding a test for just `""` would help ensure it never regresses unexpectedly.

src/index.spec.ts
```diff
@@ -1119,6 +1119,69 @@ const TESTS: Test[] = [
...
+      ["/test", ["/test", "test"]],
+      ["/abc/xyz", ["/abc/xyz", "abc/xyz"]],
+    ],
+    [],
+  ],
+  [
+    "/test/*?",
```
blakeembrey, line 1145: I think I understand why this would work, but do we want it? Or is it better/less confusing for `*` to just be its own concept? E.g. `*?` wouldn't be allowed at all, `*` is just a regex `.*` (maybe even just non-matching if the backward compatibility doesn't care about capturing it?)

HansBrende: @blakeembrey this is exactly what I enabled in my PR fyi, (making wildcard optional by default) however, non-matching I think would be a bad idea because routers, for example, might want display what page was "not found".

blakeembrey: That makes sense, but if you really want to capture it you still could by doing a capturing group like (.*), it’s not all or nothing. I’m mostly a little iffy on the idea of conflating * inheriting any parameter behavior, but I will think more it. I think it mostly comes down to what we’re trying to add asterisk support for - is it backward compatibility with existing path matching patterns or does it have other features.

wanderview: Unfortunately we are a bit locked into the behavior with urlpattern now since its shipped in two implementations.  Obviously path-to-regexp can make its own decision, but it would mean diverging at least for now.

My argument for why I think this behavior is good is that it treats the implicit prefixes and modifiers consistently for all types of matching groups.  It is also simple to explain that a `*` on its own is an alias for `(.*)`.  I feel like other options lead to more complex mental models.  (I know that @HansBrende and I disagree on this, though.)

HansBrende: @wanderview I agree, that it is indeed unfortunate that you are a bit locked in... but IMO, it will be even more unfortunate when you are _completely_ locked in once the feature comes out of experimental status and ships to _all_ major browsers. To put it another way: you are _less_ locked in **right now** than you will ever be in the future. So... all I'm saying is, let's not be too hasty and make sure the path forward is the one that makes the most sense... because the decisions made right now will affect all developers for eternity... and that is no small thing 😅

HansBrende: @wanderview re:
> It is also simple to explain that a `*` on its own is an alias for `(.*)`

I'd argue that, sure: within the constraints of "on its own", the semantics make sense. However, the logic for "on its own" is _not_ consistent (in fact, [quite contradictory](https://github.com/WICG/urlpattern/issues/163)) with the logic for "_not_ on its own"!

Consider for a moment an alternative paradigm:

_Regardless_ of the context in which the asterisk is found (whether modifying a named _or_ unnamed variable), its _auxiliary function_ is simply to change the **_default_** pattern for that variable to `.*`.

I'd say that's pretty simple to explain. Case in point ⤴

**And**, now, suddenly, there is no longer any disconnect between `*`-as-wildcard and `*`-as-quantifier! The library no longer has to deal with two contradictory paradigms, because both paradigms finally agree on a common intersection of behavior! And, _most importantly_, the 99.9999% use-case is finally achieved without breaking on edge cases.

blakeembrey: > its auxiliary function is simply to change the default pattern for that variable to .*.

That’s not it’s function though. That’d break with the behavior for + and is incompatible with implementing your own matching group definition.

blakeembrey: > it treats the implicit prefixes and modifiers consistently for all types of matching groups

This makes sense, but does it need to be a matching group. For the URLPattern use-case it’s backward compatibility with existing patterns, but do existing patterns use the match for anything? 

Another way this can be explained simply is how’s it’s already been used and how it works in regex. It’s “zero or more of the pattern”. If there’s no pattern, it can assume it is zero or more of anything, so .*.

HansBrende: @blakeembrey I'm gonna pull your comment into my own PR & respond there so as not to make this PR noisier.

wanderview: > This makes sense, but does it need to be a matching group? For the URLPattern use-case it’s backward compatibility with existing patterns, but do existing patterns use the match for anything?

> Another way this can be explained simply is how it’s already been used and how it works in regex. It’s “zero or more of the pattern”. If there’s no pattern, it can assume it is zero or more of anything, so .*.

There is no other way to do this kind of matching in path-to-regexp without producing a capture group AFAICT.  I took this as a signal that developers liked to have access to what was matched in general.  And it does seems likely there will  be at least some cases where developers will want to know what matched.  So it does seem to have some utility to me and devs who don't care can ignore that capture group output.

Sorry I didn't realize you had these concerns with the feature before we launched it on the urlpattern side.

blakeembrey:
`*` has been added back in v7 and is only allowed outside of parameter positions (not inside `{}`). It behaves like `(.*)` when matching, but splits on `/` when decoding.

I went a more extreme route to avoid complexity and make the paths easier to understand so modifiers are only valid _after_ params wrapped in `{}`. `:foo*` throws an error since it's probably an error (i.e. they can write `{:foo}*` or `*` to be explicit).

Issue #236, opened 2020-12-22
Why is pathToRegexp mutating keys parameter?


mihkeleidast:
The examples are unclear on why it works that way.
Mutating params feels wrong, because in the code it is not very easy to understand how my empty keys array gets populated. So it's always necessary to add a comment about how it works (if I want the next developer to understand it as well).

Wouldn't it make better sense to return the keys from pathToRegexp? For example, like so:
```
type pathToRegexp = (path, opts) => { matcher: Regexp, keys: Keys };
```

blakeembrey:
This is just the API that was originally designed, and I’ve avoided changing it so far. I’m not sure there’s a good reason to change it. I’ve instead moved some functionality away from raw regexes to make it a bit more user friendly. We could probably start returning `keys` as a property of the `RegExp` to continue maintaining backward compatibility.

mihkeleidast:
Immutability and clarity in code would be pretty good reasons in my opinion :)

I believe `keys` was a property of the RegExp sometime in the past. I had to figure out what happened to the keys in frctl/fractal#714 when I updated path-to-regexp from 1.7.0 to latest, hence the confusion and this issue. So reintroducing that solution would make sense to me.

nickyu42:
Hi! I'm new to the project but I wanted to contribute to it since I really like it. I was wondering if there is any interest in this issue since I wanted to try to pick it up.

blakeembrey:
This has been changed in v7 back to returning `keys` on the returned regex.

PR #272, opened 2022-03-08
Add back `keys` property to regex

nickyu42:
See issue #236

The returned regex object from `pathToRegex` used to include the modified
keys array as a `keys` property in v1.7.0. The current API is not entirely
ideal, since a keys array must be passed, we could instead use a more
functional API without mutation.

In the future this should be changed that no array needs to be passed to `pathToRegex`, 
adding it as a property this way is done to maintain backwards compatibility.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -381,6 +381,13 @@ export function match<P extends object = object>(
   return regexpToFunction<P>(re, keys, options);
 }
 
+export interface RegExpWithKeys extends RegExp {
+  /**
+   * Optional property that includes key metadata corresponding to this regex.
+   */
+  keys?: Key[];
+}
+
 /**
  * Create a path match function from `path-to-regexp` output.
  */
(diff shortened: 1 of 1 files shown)


src/index.ts
```diff
@@ -381,6 +381,13 @@ export function match<P extends object = object>(
   return regexpToFunction<P>(re, keys, options);
 }
 
+export interface RegExpWithKeys extends RegExp {
+  /**
+   * Optional property that includes key metadata corresponding to this regex.
+   */
+  keys?: Key[];
```
blakeembrey, line 388: Can we make this required? A way to get around TypeScript complaining is to use `Object.assign(regex, { keys })`.

blakeembrey:
We'll also need a simple couple of tests to ensure we don't regress but otherwise LGTM 😄

blakeembrey:
This has been done for v7, thanks!

PR #286, opened 2022-10-23
More thoroughly check for unbalanced patterns.

Lord-Kamina:
I initially had a much more complex idea, using character classes and what not (to allow `()` in some cases), but then realized that it can be easily done by how prefixes and suffixes are implemented, using `{}`


blakeembrey:
I'm sorry, I should have landed this in the last major release, but it'll be in the next one. Cheers!

PR #301, opened 2024-02-23
Add support for named matching groups.

NorbertSzydlik:
Express 4.x crashes when using named capturing group in regexp (see https://github.com/expressjs/express/issues/4277 ).
As far as I know, this feature have been added in newer version of path-to-regexp. It seems that it will land in Express 5.
This change aims to add the feature in current version of Express, which still uses path-to-regexp 0.1.x.

### DIFF
--- a/index.js
+++ b/index.js
@@ -8,6 +8,7 @@ module.exports = pathtoRegexp;
  * Match matching groups in a regular expression.
  */
 var MATCHING_GROUP_REGEXP = /\((?!\?)/g;
+var MATCHING_NAMED_GROUP_REGEXP = /\(\?<(.+?)>/g;
 
 /**
  * Normalize the given path string,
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -8,6 +8,7 @@ module.exports = pathtoRegexp;
  * Match matching groups in a regular expression.
  */
 var MATCHING_GROUP_REGEXP = /\((?!\?)/g;
+var MATCHING_NAMED_GROUP_REGEXP = /\(\?<(.+?)>/g;
```
blakeembrey, line 11: LGTM, should we just take the whole thing from the latest version? https://github.com/pillarjs/path-to-regexp/blob/c7ec332e87d8560673884d5629e1cb23cb03cb87/src/index.ts#L454-L468

NorbertSzydlik: Good point. I combined both regex'es like in latest version.

blakeembrey:
Published as `0.1.8` although it could be a breaking change if people on the previous version were relying on it _not matching_ in some way. The names do not change, only the index in the array.

PR #303, opened 2024-04-30
ci: add node versions from 17 to 22

inigomarquinez:
This PR adds node versions from 17 to 22 to the CI so the tests are run for all the different nodejs versions apart from version 16.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -19,11 +19,11 @@ jobs:
           - "21"
           - "22"
     steps:
-      - uses: actions/checkout@v2
-      - uses: actions/setup-node@v1
+      - uses: actions/checkout@v4
+      - uses: actions/setup-node@v4
         with:
           node-version: ${{ matrix.node-version }}
-      - uses: actions/cache@v2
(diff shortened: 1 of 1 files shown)


UlisesGascon (approved):
LGTM!

blakeembrey:
Why is this useful? The library doesn’t use any node.js features so testing every version seems like overkill. And I’ll have to manually increment the latest version now?

inigomarquinez:
> Why is this useful? The library doesn’t use any node.js features so testing every version seems like overkill. And I’ll have to manually increment the latest version now?

I thought this modification would be convenient for consistency with other repositories in the organisation, in which the ci runs on an array of node versions from the original to the latest active one (see [this](https://github.com/pillarjs/send/pull/229/files) or [this](https://github.com/pillarjs/finalhandler/pull/48/files)), so that you can easily verify that there is no problem with any of the nodejs versions.

Anyway, feel free to discard the PR if you don't find it interesting.

blakeembrey:
Happy to accept all the other version bumps, but I'd prefer to keep the node.js versions as-is - min version + max version tests only.

PR #314, opened 2024-08-22
update repo url

ctcpip:


UlisesGascon (approved):

blakeembrey:
Thanks! Heads up that I won’t be publishing with just this and it only affects 0.x, so it may not get released 🙏

ctcpip:
@blakeembrey yeah that makes sense, this is more out of caution.  

but it currect affects the npm page:  https://www.npmjs.com/package/path-to-regexp

that will get fixed with any release though

Jacksonearl2468 (approved):

PR #320, opened 2024-09-10
Add backtrack protection to 1.x release

blakeembrey:
I'm confident this shouldn't break 99.9% of usages, but may impact some edge cases of users of the library. Fixes ReDoS vector on matching. Closes https://github.com/pillarjs/path-to-regexp/issues/318. Does not fix ReDoS if user provides a vulnerable regex themselves, so I'll update the advisory to make it clear that it's possible to create a ReDoS if you override parameters with a custom capture and that isn't covered by the fix.

### DIFF
--- a/index.js
+++ b/index.js
@@ -72,8 +72,9 @@ function parse (str, options) {
     var partial = prefix != null && next != null && next !== prefix
     var repeat = modifier === '+' || modifier === '*'
     var optional = modifier === '?' || modifier === '*'
-    var delimiter = res[2] || defaultDelimiter
+    var delimiter = prefix || defaultDelimiter
     var pattern = capture || group
+    var prevText = prefix || (typeof tokens[tokens.length - 1] === 'string' ? tokens[tokens.length - 1] : '')
 
     tokens.push({
       name: name || key++,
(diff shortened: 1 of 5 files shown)


beaneyd-ELS:
Thank you for the fix, can you get the affected versions updated on: https://github.com/advisories/GHSA-9wv6-86v2-598j please? As 1.9 is detected as broken when it is between 0.2.0 and 8.0.0

oFlo193o:
@blakeembrey is it save to update form 0.2.5 -> 1.9.0 or are there any breaking changes, as 0.2.5 is still being used bei @nestjs/serve-static
https://github.com/nestjs/serve-static/pull/1454

matei4adrian:
Hi, @blakeembrey! This version is still seen as a vulnerable version ([CVE-2024-45296](https://nvd.nist.gov/vuln/detail/CVE-2024-45296)). The next version that is not vulnerable is 8.0.0, but this update includes breaking changes that could not be solved for packages like react-router v5. Is it possible to completely remove the vulnerability for version 1.x.x and other major versions below 8.x.x? Thanks!

PR #325, opened 2024-09-12
Update index.js (ensure path is a string before replacement)

NewEraCracker:
Fixes: https://github.com/expressjs/express/issues/5955

Ensure the `path` variable is explicitly converted to a string before applying regex replacements to prevent potential errors with non-string inputs.

This change enhances robustness and correctness in handling various input types.

### DIFF
--- a/index.js
+++ b/index.js
@@ -65,7 +65,7 @@ function pathToRegexp(path, keys, options) {
     return new RegExp(path.join('|'), flags);
   }
 
-  path = path.replace(
+  path = String(path).replace(
     /\\.|(\/)?(\.)?:(\w+)(\(.*?\))?(\*)?(\?)?|[.*]|\/\(/g,
     function (match, slash, format, key, capture, star, optional, offset) {
       pos = offset + match.length;


NewEraCracker:
It was being silently coerced to string before:
[`path = ('^' + path + (strict ? '' : path[path.length - 1] === '/' ? '?' : '/?'))`](https://github.com/pillarjs/path-to-regexp/commit/29b96b4a1de52824e1ca0f49a701183cc4ed476f#diff-e727e4bdf3657fd1d798edcd6b099d6e092f8573cba266154583a746bba0f346L64)

Feel free to think about other fixes to avoid crash.

alewitt2:
i guess the question is what's the desired behavior then. is it acceptable to return undefined for path if its not a string?

blakeembrey:
I'm going to close this out. Although the library previously accepted non-strings accidentally here, it wasn't intentional or a guaranteed part of the API, and it looks like it may have actually hidden real bugs from users and upstream libraries like Express. Thanks for the attempt at patching though!

blakeembrey:
The 0.1.x branch is fairly clear in the README https://github.com/pillarjs/path-to-regexp/tree/0.1.x?tab=readme-ov-file#pathtoregexppath-keys-options

> A string in the express format, an array of such strings, or a regular expression

az-faro:
Maybe you could throw an exception saying it "has to be a string, an array of strings or a regex" if it's neither of them? That way it's crystal clear if you use the library incorrectly. Whereas as it is now, one could mistake this for a bug in the library instead of a mistake on your own side.

PR #352, opened 2025-02-25
Test CI with both Node.js and Bun

Electroid:
This adds support to run tests in CI using Bun, and keeps Node.js support. It makes running tests 35% faster.

<table>
<tr>
<td align="center" width="50%"><strong>Bun</strong> - 11s</td>
<td align="center" width="50%"><strong>Node</strong> - 17s</td>
</tr>
<tr>
<td align="center" valign="top" width="50%">Screenshot 2025-02-20 at 3 32 02 PM</td>
<td align="center" valign="top" width="50%">Screenshot 2025-02-20 at 3 32 49 PM</td>
</tr>
</table>

Since Bun supports Jest & Vitest, you'll notice there are no code changes! It just works. (and I did test the Github Action on a fork to verify that it indeed does work)

This also changes `npm install` to `bun install`, but just in CI. Developers can continue to use npm, yarn, etc.

I also updated various GitHub Action steps to use the newer versions (e.g. `actions/checkout@v2` -> `actions/checkout@v4`)

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -10,8 +10,12 @@
 
 ## Installation
 
-```
-npm install path-to-regexp --save
+Install with your choice of package manager: npm, yarn, or bun.
+
+```sh
+npm install path-to-regexp
+yarn add path-to-regexp
+bun add path-to-regexp
 ```
 
 ## Usage
@@ -25,6 +29,8 @@ const {
   stringify,
 } = require("path-to-regexp");
 ```
+
+This library is compatible and tested with Node.js 16+ and Bun 1.2+.
 
 ### Parameters
 
--- a/bunfig.toml
+++ b/bunfig.toml
@@ -1,0 +1,3 @@
+[test]
+coverage = true
+coverageReporter = ["text", "lcov"]
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,27 +5,42 @@ on:
 permissions:
   contents: read
 jobs:
-  test:
-    name: Node.js ${{ matrix.node-version }}
+  test-node:
+    name: Node.js (${{ matrix.node-version }})
     runs-on: ubuntu-latest
     strategy:
       matrix:
         node-version:
           - "16"
-          - "*"
+          - "latest"
     steps:
-      - uses: actions/checkout@v2
-      - uses: actions/setup-node@v1
+      - uses: actions/checkout@v4
+      - uses: actions/setup-node@v4
         with:
           node-version: ${{ matrix.node-version }}
-      - uses: actions/cache@v2
+      - uses: oven-sh/setup-bun@v2
         with:
-          path: ~/.npm
-          key: ${{ runner.os }}-node-${{ hashFiles('**/package-lock.json') }}
-          restore-keys: |
-            ${{ runner.os }}-node-
-      - run: npm ci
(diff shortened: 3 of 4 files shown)


Readme.md
```diff
@@ -26,6 +30,8 @@ const {
 } = require("path-to-regexp");
 ```
 
+This library is compatible and tested with Node.js 16+ and Bun 1.2+.
```
blakeembrey, line 33: I'd prefer to remove this, it's all just JS so it's compatible with anything that supports ES2015 basically. These just happen to be where tests are run.

bunfig.toml
```diff
@@ -0,0 +1,3 @@
+[test]
+coverage = true
+coverageReporter = ["text", "lcov"]
```
blakeembrey, line 3: Is this needed?

blakeembrey (approved):
@pillarjs/express-tc This project is JS-only, no deps, nothing runtime specific (since it's designed to be used everywhere). Based on past discussions we would want to merge this (and any other runtimes people want to test on) but my personal experience is to keep the test suite minimal and instead target ES version supports vs runtimes.

blakeembrey:
> why?

@ctcpip when we’ve discussed it in the past I thought you were on the side of testing everything that’s supported. The package supports everything that can run ES2015, including browsers, Cloudflare workers, etc so it seems like this is an improvement toward that goal.

ctcpip:
> > why?
> 
> @ctcpip when we’ve discussed it in the past I thought you were on the side of testing everything that’s supported. The package supports everything that can run ES2015, including browsers, Cloudflare workers, etc so it seems like this is an improvement toward that goal.

That's not exactly my position.  My position is that if we are **officially and deliberately** supporting something, then we should have tests for those runtimes, environments, engines, etc.

Similarly, I wouldn't expect to see tests for every JS engine out there either.

The [ADR](https://github.com/expressjs/discussions/pull/323/files) on this is not yet merged, but right now contains the following:

> It is also worth noting that Express was specifically designed to run with Node.js. While some of our libraries can run in other runtimes, this can currently be classified as an "unofficial and untested feature." Consequently, our CI systems do not include other runtimes as part of their testing workflows.

jonchurch:
Thank you for the contribution. 

Im +1 to keeping a minimal test suite

Is there a specific problem this PR would solve for consumers or  maintainers?

If so, Im open to learning more.

Absent a clear problem, Im not seeing value in the additional workflow complexity

blakeembrey:
@ctcpip I did try my best to get that language removed with https://github.com/expressjs/discussions/pull/323#discussion_r1929031726 🤷 This package does support other runtimes than it's tested on, but that is due to redundancy and maintenance overhead.

@Electroid I'm going to close this out for now since this is a simple package and supports any ES2015 environment, so it shouldn't need a dedicated test suite for this. Let me know if you disagree though. This is currently enforced through TypeScript with [`target`](https://github.com/pillarjs/path-to-regexp/blob/19c5659d67bb2c3f74d65bab3b563c15ab3fd79c/tsconfig.json#L4-L5) and [`types`](https://github.com/pillarjs/path-to-regexp/blob/19c5659d67bb2c3f74d65bab3b563c15ab3fd79c/tsconfig.build.json#L4) being explicitly defined.

PR #355, opened 2025-04-07
Add tsbuildinfo to .gitignore

simhnna:


bjohansebas:
When is that file generated?

simhnna:
I ran `npm test` locally. It's a typescript file that is generated if `incremental` is set in tsconfig and is used to speed up builds. 
https://www.typescriptlang.org/tsconfig/#incremental

bjohansebas (approved):

Issue #336, opened 2024-10-17
Improve error message when parsing fails


simhnna:
I got error messages like `TypeError: Missing parameter name at 3: https://git.new/pathToRegexpError` and I was wondering how that weird url made it's way into my code

Two suggestions:
1. It doeesn't look like `lexer()` gets the whole error, so maybe it would be good to wrap those errors in `parse()` and include the whole string?
2. Make it clear that the url is documentation. Something like `For more information about regexp paths check https://...`
3. I personally would prefer a link that clearly shows it's going to github instead of a link shortener, but that's just my opinion

If you're ok with me changing 1 and 2, I'm happy to prepare a pull request

blakeembrey:
1. I'm not sure I understand, can you elaborate on what "the whole string"?
2. Sounds good, I'm open to a PR that changes it to something like `...; visit <link> for more information`.
3. Prefer to avoid changing for now, it's there to ensure the error docs can move or update in the future without breaking old versions of the library.

abs-lme:
For 1, I guess the problem is that the error message gives no indication on where the error occurred. I have the same error stating something about 'missing parameter name at 2' without any details. 
So I have to check all my routes for potential errors.
Dumping the input path will already help with handling this error.

simhnna:
I'm sorry I completely forgot about that. I created two merge requests that address points 1 and 2

SlothScript:
I have created a simple version of this that shows you the problematic url. pull request #358

PR #356, opened 2025-04-07
Improve error messages by including the path

simhnna:
Without the path users need to search all their code for a malformed path or debug to find the issue

Closes: #336

### DIFF
--- a/src/index.spec.ts
+++ b/src/index.spec.ts
@@ -15,14 +15,14 @@ describe("path-to-regexp", () => {
     it("should throw on unbalanced group", () => {
       expect(() => parse("/{:foo,")).toThrow(
         new TypeError(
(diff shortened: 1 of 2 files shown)


simhnna:
Tried a couple of different places to include the path in the error message and I think prepending it is the easiest to read but if you'd like it somewhere else I'd update accordingly

blakeembrey (commented):
I appreciate this PR, it's a great direction! However I think it'd be good to avoid using `stringify` as it can output a different path to the original input, and instead add a `originalPath` or something to `TokenData`. I can open a PR you can review if that's OK?

blakeembrey:
I've just merged https://github.com/pillarjs/path-to-regexp/pull/363 which adds the original path to the error.

PR #358, opened 2025-04-15
Show problematic url

SlothScript:
Show the URL that is giving the error to help debugging instead of just giving the debug url.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -145,12 +145,12 @@ function* lexer(str: string): Generator<LexToken, LexToken> {
       }
 
       if (pos) {
-        throw new TypeError(`Unterminated quote at ${pos}: ${DEBUG_URL}`);
+        throw new TypeError(`Unterminated quote at ${pos}: ${str}; Visit ${DEBUG_URL} for more information.`);
       }
     }
 
     if (!value) {
-      throw new TypeError(`Missing parameter name at ${i}: ${DEBUG_URL}`);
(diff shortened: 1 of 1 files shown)


SlothScript:
Sorry, forgot to mention issue #336

src/index.ts
```diff
@@ -203,7 +203,7 @@ class Iter {
     if (value !== undefined) return value;
     const { type: nextType, index } = this.peek();
     throw new TypeError(
-      `Unexpected ${nextType} at ${index}, expected ${type}: ${DEBUG_URL}`,
+      `Unexpected ${nextType} at ${index}, expected ${type}: ${str}; Visit ${DEBUG_URL} for more information.`,
```
blakeembrey, line 206: I don't think we have access to `str` in this method.

src/index.ts
```diff
@@ -575,7 +575,7 @@ function toRegExp(tokens: Flattened[], delimiter: string, keys: Keys) {
 
     if (token.type === "param" || token.type === "wildcard") {
       if (!isSafeSegmentParam && !backtrack) {
-        throw new TypeError(`Missing text after "${token.name}": ${DEBUG_URL}`);
+        throw new TypeError(`Missing text after "${token.name}": ${str}; Visit ${DEBUG_URL} for more information.`);
```
blakeembrey, line 578: No original `str` here either. We need to make some changes to pass it around as some methods support just accepting `TokenData` only.

blakeembrey:
I rolled something similar into the PR here that I've just merged: https://github.com/pillarjs/path-to-regexp/pull/363

Issue #223, opened 2020-05-01
Exact Match by non capturing group but allow wildcard for subsegments


janus-reith:
I'm trying to rewrite non-localized paths (e.g /page1) to localized ones (/de/page1) and therefore I'm seeking help to get the following to work:

- "/de" or "/en" should NOT be matched
- "/de/anything" or "/en/anything" should NOT be matched

- "/test" or "/test/anything" should be matched
- "/deno" or "/end" should be matched

With my current implementation which is based on: https://github.com/pillarjs/path-to-regexp/blob/0c466b1b0944e8d0022b5b15069364a8483bf9c5/src/index.spec.ts#L2498

I can get the first 3 of 4 requierements, but can't prevent "de" from also matching "deno".
It seems like instead of allowing any chars which are not "/" after the lang segment, I would need a solution which allows either an exact match or a wilcard which must be preceded by another "/". (So both /de and /de/anything would be covered).

Segmenting these values any further is not a requierement as I'm taking `req.url` as a whole and only prepend the locale.

This sandbox should probably make it easy to get what I mean:
https://codesandbox.io/s/path-to-regexp-demo-c9efk?file=/src/index.js
(Result is in the console)

I got a bit stuck here, and any help is appreciated, thank you! :)

blakeembrey:
This is pretty complex and a little outside of how `path-to-regexp` is intended to be used. My suggestion would be to split this into separate routes and have them ordered. Do you know what framework you're using this with? Depending on where, you could just match `/:lang` first, and if it didn't match your expected language pass it to child routes.

janus-reith:
@blakeembrey Thanks for your help.
I'm using nextjs, which allows for an ordered list of matches, and pass those that don't match to the next one. (Functionality is described here: https://github.com/zeit/next.js/issues/9081)

Regarding separate routes, yes I also that idea, but thought that having one expression to match all cases would've been cleaner and I'm just to unskilled with regexpes to get this right.
But I'm happy to put this into separate routes if that is easier.

So, with separate routes, would I have
1. one that does exact matches like /abc or /test or /deno, but does not match /de or /en
2. one that does exact matches but allows for a wilcard after a "/" occured, so it matches /abc/[...] and /test/[...], but not /de[...] and /en[...]?

[...]

PR #362, opened 2025-05-10
feat: support parameter patterns

Kakulukian:
I'm planning to move to the newer version of path-to-regexp. In our codebase heavily relies on regex for parameter validation, especially in some legacy components we still have running in production.

I stumbled across [this comment](https://github.com/pillarjs/path-to-regexp/issues/223#issuecomment-2386812589) that mentions parameter pattern support might be coming in future versions. Based on that, I went ahead and implemented the functionality in this PR.

Feel free to ignore this PR if it doesn't align with the project's direction. I understand you may have different plans for handling parameter patterns in future versions!

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -3,8 +3,23 @@ const NOOP_VALUE = (value: string) => value;
 const ID_START = /^[$_\p{ID_Start}]$/u;
 const ID_CONTINUE = /^[$\u200c\u200d\p{ID_Continue}]$/u;
 const DEBUG_URL = "https://git.new/pathToRegexpError";
-const INVALID_PATTERN_CHARS = "^$.+*?[]{}\\^";
-
+const PATTERN_META_CHARS = new Set([
+  "*",
+  "+",
+  "?",
+  ".",
+  "^",
+  "$",
+  "|",
+  "\\",
+  "(",
+  ")",
+  "[",
+  "]",
+  "{",
+  "}",
+]);
+const ALLOWED_PATTERN_META_CHARS = new Set(["|"]);
 /**
  * Encode a string into another string.
  */
@@ -166,12 +181,6 @@ function* lexer(str: string): Generator<LexToken, LexToken> {
     while (i < chars.length && depth > 0) {
       const char = chars[i];
 
-      if (INVALID_PATTERN_CHARS.includes(char)) {
-        throw new TypeError(
-          `Only "|" is allowed as a special character in patterns at ${i}: ${DEBUG_URL}`,
-        );
-      }
-
       if (char === ")") {
         depth--;
         if (depth === 0) {
@@ -284,6 +293,7 @@ export interface Parameter {
 export interface Wildcard {
   type: "wildcard";
   name: string;
+  pattern?: string;
 }
 
 /**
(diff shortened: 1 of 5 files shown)


blakeembrey:
@Kakulukian I wanted to let you know ASAP that I really appreciate the PR! It is the right direction, but unfortunately it's even more involved than this because I want to add it back in a way that ensures it still generates safe regexes. That sadly means more of a regex parser.

The good news, if you're interested, is that we could start with a very simple "parser" by rejecting all meta characters except `|`. Then add other use-cases one by one. Would that work for you?

Assuming you're up for it, I'll also separately leave some review comments.

src/index.ts
```diff
@@ -81,7 +81,7 @@ type TokenType =
 interface LexToken {
   type: TokenType;
   index: number;
-  value: string;
+  value: string | { name: string; pattern?: string };
```
blakeembrey, line 84: Instead of this, it'd be simpler to add `PATTERN` to `TokenType`. The `value` of the pattern would be what is between the two `()`.

src/index.ts
```diff
@@ -153,7 +156,29 @@ function* lexer(str: string): Generator<LexToken, LexToken> {
       throw new TypeError(`Missing parameter name at ${i}: ${DEBUG_URL}`);
     }
 
-    return value;
+    if (chars[i] === "(" && options?.pattern) {
```
blakeembrey, line 159: Take a look at the previous code here: https://github.com/pillarjs/path-to-regexp/commit/60f2121e9b66b7b622cc01080df0aabda9eedee6#diff-a2a171449d862fe29692ce031981047d7ab755ae7f84c707aef80701b3ea0c80L155

I'd use that as a starting point, e.g. move the `(` into the `lexer` itself (not part of `name`) and yield a `PATTERN` directly. Handle the fact that `PATTERN` is expected in a certain position in the `if (param)` below by doing `tryConsume("PATTERN")`.

Kakulukian:
> @Kakulukian I wanted to let you know ASAP that I really appreciate the PR! It is the right direction, but unfortunately it's even more involved than this because I want to add it back in a way that ensures it still generates safe regexes. That sadly means more of a regex parser.
> 
> The good news, if you're interested, is that we could start with a very simple "parser" by rejecting all meta characters except `|`. Then add other use-cases one by one. Would that work for you?
> 
> Assuming you're up for it, I'll also separately leave some review comments.

That's ok for me, updated with your suggestions :)

blakeembrey (commented):
Feel free to omit the validation in this PR, and either of us can work on it in a follow up PR.

rokob:
What is the status here? Would you be open to someone taking this over and moving it forward or a new PR doing something similar? The lack of this feature is blocking my ability to upgrade other libraries that depend on v8 here. I'd be happy to move this forward. I need alteration `|` and patterns like `:id([a-z]{6})` but that seems inline with that general approach here.

blakeembrey:
@rokob For now I've open a PR with trivial pattern support: https://github.com/pillarjs/path-to-regexp/pull/417. If we use that as a starting point it might make it easier for others to contribute so there's less comments on integrating it with everything and it can focus on the validation/pattern support itself.

PR #395, opened 2025-08-07
Reduce bytes in parse function

blakeembrey:
Shaves ~2.5% (1.95kb -> 1.9kb) of build output size by further simplifying the merged code between the lexer and parser started in https://github.com/pillarjs/path-to-regexp/pull/363.


ctcpip (approved):
looks like this improves perf a bit as well

UlisesGascon (approved):

blakeembrey:
> looks like this improves perf a bit as well

Just a tiny bit, but hopefully no one is parsing that much to matter anyway.

PR #396, opened 2025-08-07
100% test coverage

blakeembrey:
Fixing up 100% test coverage cases, additionally shifting some bytes around for readability and size (mainly the `.map` to a `for` loop to make the recursive case more obvious).

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -487,7 +487,7 @@ export function pathToRegexp(
   for (const input of pathsToArray(path, [])) {
     const data = typeof input === "object" ? input : parse(input, options);
     for (const tokens of flatten(data.tokens, 0, [])) {
-      sources.push(toRegExp(tokens, delimiter, keys, data.originalPath));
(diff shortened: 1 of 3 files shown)


src/index.ts
```diff
@@ -619,40 +619,65 @@ function negate(delimiter: string, backtrack: string) {
...
-    .map(function stringifyToken(token, index, tokens): string {
-      if (token.type === "text") return escapeText(token.value);
-      if (token.type === "group") {
-        return `{${token.tokens.map(stringifyToken).join("")}}`;
-      }
+function stringifyTokens(tokens: Token[]): string {
+  let value = "";
+  let i = 0;
+
+  function name(value: string) {
```
blakeembrey, line 628: Pulled `name` function out (which adds bytes) because it's needed to avoid a crash on `name` being missing _before_ the token type error.

blakeembrey:
CleanShot 2025-08-07 at 15 21 19@2x

jonchurch:
Requested myself so I can find it again if I dont get to it after the tc meeting

Issue #346, opened 2025-01-20
Please add ESM export to v8


zoxon:
Have a problem with vite and typescript when I try to use current CJS build ([ERROR] exports is not defined)

https://github.com/microsoft/TypeScript/issues/32934

blakeembrey:
Can you share the full error you’re getting? I have used it successfully with Vite and that linked issue doesn’t appear to be related.

zoxon:
I am using it together with Astro. Astro uses vite for builds

Here is part of the logs related to the error

```sh
  vite:resolve 0.20ms path-to-regexp -> /Users/zoxon/work/public-sites/node_modules/path-to-regexp/dist/index.js +1ms
  vite:resolve 0.20ms ufo -> /Users/zoxon/work/public-sites/node_modules/ufo/dist/index.mjs +0ms
  vite:import-analysis 0.40ms [6 imports rewritten] src/core/routes/helpers.ts +2ms
  vite:transform 1.04ms /src/core/routes/helpers.ts +2ms
  vite:load 0.11ms [fs] ../../node_modules/path-to-regexp/dist/index.js +2ms
  vite:import-analysis 0.07ms [no imports] /Users/zoxon/work/public-sites/node_modules/path-to-regexp/dist/index.js +1ms
  vite:transform 0.15ms [skipped] ../../node_modules/path-to-regexp/dist/index.js +1ms
  astro:vite Error when evaluating SSR module /Users/zoxon/app/src/pages/[...region]/[...lang]/market/[...ticker].astro:
  astro:vite |- ReferenceError: exports is not defined
  astro:vite     at eval (/Users/zoxon/work/public-sites/node_modules/path-to-regexp/dist/index.js:4:23)
  astro:vite     at ESModulesEvaluator.runInlinedModule (file:///Users/zoxon/work/public-sites/node_modules/astro/node_modules/vite/dist/node/module-runner.js:1057:6)
  astro:vite     at SSRCompatModuleRunner.directRequest (file:///Users/zoxon/work/public-sites/node_modules/astro/node_modules/vite/dist/node/module-runner.js:1271:82)
  astro:vite     at SSRCompatModuleRunner.directRequest (file:///Users/zoxon/work/public-sites/node_modules/astro/node_modules/vite/dist/node/chunks/dep-BJP6rrE_.js:30789:35)
[...]

Issue #347, opened 2025-02-03
v7 and v8 are not tree-shakable


samuelstroschein:
## Problem 

Tree-shaking is not working with v7 and above because the ESM build has been removed. 

I am the author of [Paraglide JS](https://inlang.com/m/gerre34r/library-inlang-paraglideJs). Tree-shaking is a key differentiator. For locale detection reasons, I'd like to use path-to-regexp on the client. The lack of tree-shaking (which would increase the runtime size from <1kb to >4kb) and ESM problems like #346 prevent me from adopting path-to-regexp. 

Using `URLPattern` instead of path-to-regexp is not possible because a reverse compile function is missing https://github.com/whatwg/urlpattern/issues/73.  

### Reproduction 

1. Open https://bundlejs.com/?q=path-to-regexp%407&treeshake=%5B%7B+compile+%6D%5D
2. Press `build`
3. Repeat with v7 and v8

To inspect the bundle, disable `minify: true` in the settings. V6 only contains functions used in `compile,` while >V7 bundles everything. 

## Proposal 

Re-introduce dual CJS + ESM build, [or say goodbye to CJS altogether, now that Node supports native `require(esm)`](https://nodejs.org/en/blog/announcements/v22-release-announce#support-requireing-synchronous-esm-graphs).

bjohansebas:
Sorry, but at the moment we are not going to release the package with esm, see https://github.com/expressjs/discussions/pull/323

wesleytodd:
I thought @blakeembrey said he was planning on supporting a dual most package? Maybe I misunderstood our discussion yesterday, but I think we might want to leave issues like this open for discussion to ensure we get feedback and have time to understand the issues folks are reporting.

In this case, I would love to know what benefit you get out of treeshaking a single file dependency? The file is 14kb unminified, and nearly all of it would be used together in all the use cases I have seen. So what is the end user benefit of treeshaking for such a package?

ljharb:
Also, CJS tree-shakes just fine (albeit often with an extra plugin)

samuelstroschein:
> @wesleytodd  I would love to know what benefit you get out of treeshaking a single file dependency? The file is 14kb unminified, and nearly all of it would be used together in all the use cases I have seen. So what is the end user benefit of treeshaking for such a package?

To not ship the entire file to the client and avoid import errors of users who are using Paraglide JS. 
 
[...]

PR #397, opened 2025-08-09
Support ESM package export

re-taro:
Summary

Add ESM (ECMAScript Module) support alongside existing CommonJS compatibility. This change enables the package to be used with both import and require statements, providing better compatibility with modern JavaScript tooling and environments.

Resolves: #346, #347

Changes
- Package Configuration: Added dual package exports with proper conditional exports for ESM and CommonJS
- Build System: Implemented custom build script to generate both ESM and CommonJS outputs
- TypeScript Configuration: Split build configuration into separate files for ESM and CommonJS compilation targets
- Module Type: Set package type to "module" as the primary format while maintaining CommonJS compatibility

Technical Details

- ESM output is placed in dist/ directory as the default export
- CommonJS output is placed in dist/cjs/ directory with its own package.json specifying "type": "commonjs"
- Build script handles proper package.json generation for distribution
- Size limits updated to monitor both ESM and CommonJS bundle sizes

Breaking Changes

None. This change maintains full backward compatibility with existing CommonJS consumers while adding ESM support.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,7 +1,7 @@
 {
   "name": "path-to-regexp",
   "version": "8.2.0",
-  "lockfileVersion": 2,
+  "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
@@ -20,7 +20,11 @@
         "vitest": "^3.0.5"
       },
       "engines": {
-        "node": ">=16"
+        "node": ">=18"
+      },
+      "funding": {
+        "type": "opencollective",
+        "url": "https://opencollective.com/express"
       }
     },
     "node_modules/@ampproject/remapping": {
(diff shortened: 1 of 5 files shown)


re-taro:
@blakeembrey @bjohansebas 

Please let me know your thoughts at your earliest convenience.

UlisesGascon:
@re-taro there is a policy in the oven regarding ESM https://github.com/expressjs/discussions/pull/323. I hope that this can give you additional context :+1:

blakeembrey:
Unfortunately, as mentioned in the issues you linked to, this won't land until there's a new major. I don't believe this is backward compatible because anyone it will change the imports for anyone using ESM to import from CommonJS today.

PR #398, opened 2025-08-19
Add custom error class

blakeembrey:
Follow up from https://github.com/pillarjs/path-to-regexp/pull/363#discussion_r2283012345. Found a solution that adds `originalPath` to the errors in a way that doesn't add to stack size, increases the build by only 0.01kb (half the original test in my comment), and adds `originalPath` as a property to all the errors.


bjohansebas (approved):
LGTM!

PR #399, opened 2025-08-19
Remove engines from package

blakeembrey:
Engines was bumped in https://github.com/pillarjs/path-to-regexp/pull/348 to support the latest vitest version, removing the engines field instead to release a new minor with the changes on main. I could alternatively revert to `>=16` but I don't want it out of sync with CI or to figure out how to downgrade vitest for this.

### DIFF
--- a/package.json
+++ b/package.json
@@ -44,9 +44,6 @@
     "typescript": "^5.7.3",
     "vitest": "^3.0.5"
   },
-  "engines": {
-    "node": ">=18"
-  },
   "publishConfig": {
     "access": "public"
   },


bjohansebas (commented):
Why did it work before? What version of Vitest were they using before? Or did it always use the latest, and at some point something broke because of Vitest?

Really, whether the engines field is kept or removed, it should still respect the CI support, which was Node 16. It’s kind of unlikely for path-to-regexp to break, but not impossible

blakeembrey:
Why did what work before? Vitest was upgraded to fix a security vulnerability, the latest vitest is node 18+.

Actual node version support is pretty meaningless in this package, as there's nothing related to node used and through TypeScript checks the ES version is enforced.

bjohansebas (approved):
Knowing that helps, It’s fine by me.

UlisesGascon (approved):
LGTM! Just to clarify for those following the engines ADR discussion. The consensus seems to be not to treat removing engines as a major change (see: https://github.com/expressjs/discussions/pull/289#issuecomment-2852028266). 

This PR is aligned with that proposal :+1:

PR #402, opened 2025-08-26
Move parameter name errors up in docs

blakeembrey:
The `/*` issue seems to be the most common, move it up and trying to make it more clear. Open to feedback on phrasing or rewriting the sentences entirely.


bjohansebas (commented):
In the Express documentation, I decided to use examples for this point. I’m not even sure if that’s been helping or not, but maybe here it could have an effect and help people migrate

Image

Here you can use the diff format for those examples, showing what needs to be removed or added.

I'll leave the decision up to you.

blakeembrey:
@bjohansebas I'll play around with it, right now the docs are pretty short so I'll try expanding them all and see how it looks.

PR #404, opened 2025-09-05
Remove dependabot config

blakeembrey:
Reducing PR noise. Switching to configuration in GitHub settings for security updates using Dependabot instead of a config file.


UlisesGascon (approved):

lannathompson65-arch (commented):
Done

Issue #380, opened 2025-06-11
Parameter customization


blakeembrey:
In versions < 8 it was possible to customize the parameter match using a regex. For example, `:param(a|b|c)`. While this is immensely helpful for end users, it created a security issue if the user manages to generate a vulnerable regex. This is incredibly hard for users to avoid because the individual parameters may not be vulnerable to ReDoS but when combined together it generates something unsafe.

This issue is to bring back the same parameter match override (the part after `:param` between parens) but doesn't re-introduce a vulnerability. Two approaches I can think of are:

1. Whitelist a subset of regex features that can be safely combined
2. Add a DSL that translates into a safe regex

I think the best path forward is likely a regex-like DSL, given that regex itself can be different depending on the flags used, and there's no control over the flags in this library (e.g. unicode support). For majority of people, it would just be regex anyway.

An additional benefit of having a little more control over the customization instead of directly passing through a regex is also support for `encodePath`. In older versions, there was a disconnect because the path itself was encoded for matches but the regex was not, so something trivial like `/^` and `/:param(^)` resulted in different regexes.

Helveg:
I don't mind contributing this, what should the DSL support, and what constitutes a safe/unsafe regex?

blakeembrey:
> what should the DSL support

I think the safe place to start would be standard regex features via whitelist. We could start with just text and no regex, and `|`, then add more features such as `\d`, `\w`, etc. Once we get into repeated characters is where the unsafe regex start to matter.

> what constitutes a safe/unsafe regex

Whether it could cause backtracking. Mostly it's a question of whether this regex could mix with the previously generated text to create a catastrophic backtracking situation.

Helveg:
I have a work-in-progress [here](https://github.com/pillarjs/path-to-regexp/compare/master...Helveg:path-to-regexp:regex-subset).

I think the allowed regex features make sense:

* Only non-capturing groups allowed with `(?:`, no nested groups, no lookahead, lookbehind, etc to worry about.
* Character classes allowed
* Quantifiers allowed
* No backreferences
* No anchors

I'm just a bit confused about 2 things:

* How to add the parameter regex back into the main regex? (I currently just put it in without any checks)
* How to deal with delimiters? For example `/user/:id([^x]+)` would gobble up `/user/12345/related`, I'm not sure if that's something we're supposed to prevent somehow?

Any feedback appreciated :)

blakeembrey:
[...]

PR #417, opened 2026-01-13
Add param pattern support

blakeembrey:
Extremely simple version of https://github.com/pillarjs/path-to-regexp/issues/380, meant to be a starting point for more iteration on the actual regex allowed.

### DIFF
--- a/package.json
+++ b/package.json
@@ -50,7 +50,7 @@
   "size-limit": [
     {
       "path": "dist/index.js",
-      "limit": "2 kB"
+      "limit": "2.1 kB"
     }
   ],
   "ts-scripts": {
--- a/src/cases.spec.ts
+++ b/src/cases.spec.ts
@@ -146,6 +146,16 @@ export const PARSER_TESTS: ParserTestSet[] = [
       "\\\\:test",
     ),
   },
+  {
+    path: "/:username([a-zA-Z]+)",
+    expected: new TokenData(
+      [
+        { type: "text", value: "/" },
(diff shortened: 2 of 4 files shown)


GroophyLifefor (approved):

bjohansebas (commented):
Okay, the code looks good. Could you please update the README?

Helveg:
Is it possible to also add other common valid URL characters like `- _ ! $ & ' , ; =` to the allowed characters? They seem completely innocent. Maybe even `+` (commonly used for spaces) and `*`, `?`, I'm not sure they can do harm if `( )` are not allowed, perhaps they should only be allowed when they're explicitly escaped like `\+`, `\*`, `\?` ?

blakeembrey:
> Is it possible to also add other common valid URL characters like - _ ! $ & ' , ; = to the allowed characters?

Absolutely, this was just a scaffold to start iterating on but I can add them to the PR. One open question on `^` and `$`, since they're invalid in the regex output, is whether they should be escaped? I'm leaning toward yes and documenting the supported regex features/DSL in the README clearly.

> Maybe even + (commonly used for spaces) and *, ?, I'm not sure they can do harm if ( ) are not allowed, perhaps they should only be allowed when they're explicitly escaped like \+, \*, \? ?

Depends on if we're supporting repeat/optional features. If so, yes, they'd need to be escaped. That's where using a regex to validate completely breaks down and, with repeats, a ReDoS vector opens so it's a more complex PR.

todor-a:
hey, what can I do to help move this forward? :)

PR #423, opened 2026-03-25
Byte optimization

blakeembrey:
Heavy optimization to reduce the bytes introduced from the last couple of PRs.


blakeembrey:
Trie deduplication bumped package to 2.23 kB. Previously was 2.01 kB.

This PR reduces us a little bit back to 2.08 kB.

PR #440, opened 2026-04-02
Add ReDoS regression coverage for sequential optional groups

TheWatcher01:
Adds regression coverage for sequential optional groups that were previously implicated in ReDoS-style regex growth, plus a longer variant to keep the guard stable.

Also adds vite as a dev dependency so the local Vitest config can load during test runs.

### DIFF
--- a/package.json
+++ b/package.json
@@ -42,6 +42,7 @@
     "recheck": "^4.5.0",
     "size-limit": "^11.1.2",
     "typescript": "^5.7.3",
+    "vite": "5.4.20",
     "vitest": "^3.0.5"
   },
   "publishConfig": {
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -1,0 +1,2431 @@
+lockfileVersion: '9.0'
+
+settings:
+  autoInstallPeers: true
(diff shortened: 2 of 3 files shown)


blakeembrey:
These were never vulnerable, and they're already tested against due to the existing run of redos checks on all the match cases. I also can't accept the PR with changes to the package manager and dependencies.

blakeembrey:
If you really want to include this test, you can add it to the existing match tests and write up the match cases of each possible combination of optional parameters. But this is basically covered by existing combinations because optionals are just expanded, so what you're testing here is just that `/users/:id`, `/users/:tab`, `/users/:mode`, `/users/:id/:tab` and so on for every combination works. These simple patterns have never had a ReDoS, what the expansion issue had in the vulnerability report was a that it produced `2^n` possible matches with `2^n` performance.

Unfortunately the report was modified heavily (not by me) so it looks like this nuance, I'm going to investigate internally whether we can make it clearer.

blakeembrey:
TL;DR you should check if it was failing on the old version before adding the test, and confirm whether existing test coverage is insufficient.

PR #443, opened 2026-06-11
Avoid duplicate trailing delimiter matches

### LINKED ISSUE #441: regexp should not end with both explicit end-slash and optional end-slash
Hello and thanks for this package. When the original path ends with an end-slash (ex '/about/') the returned regexp ends with both a required and an optional end-slash; I believe, instead, it should end with only one end-slash; optional or required.

Instead of this,
```javascript
pathToRegexp('/about/').regexp // /^(?:\/about\/)(?:\/$)?$/i
```

One of these would be better,
```javascript
pathToRegexp('/about/').regexp // /^(?:\/about)(?:\/$)?$/i
pathToRegexp('/about/').regexp // /^(?:\/about\/)$/i
```

vibhor-aggr:
Fixes #441.

`pathToRegexp('/about/')` currently appends the optional trailing delimiter after a source that already ends in `/`, so the generated matcher accepts `/about//`. This moves the optional trailing delimiter into each flattened alternative and skips it when that alternative already ends with the configured delimiter.

Verification:
- `./node_modules/.bin/vitest run src/index.spec.ts -t '/test//'`
- `./node_modules/.bin/vitest run src/index.spec.ts`
- `npm run build`
- `npm run size`
- `git diff --check`

Note: I also attempted `npm test`; it reached 477 passing Vitest tests but the local ReDoS helper binary requires GLIBC_2.32/GLIBC_2.34 while this host has glibc 2.28, so that command exits with `recheck-linux-x64` EPIPE errors in this environment.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -464,16 +464,14 @@ export function pathToRegexp(
     }
 
     const data = typeof path === "object" ? path : parse(path, options);
+    const originalPath = data.originalPath;
     flatten(data.tokens, 0, [], (tokens) => {
       if (combinations >= 256) {
-        throw new PathError("Too many path combinations", data.originalPath);
+        throw new PathError("Too many path combinations", originalPath);
       }
 
       if (combinations > 0) source += "|";
-      source += toRegExpSource(tokens, delimiter, keys, data.originalPath);
-      if (trailing && !tokensEndWithDelimiter(tokens, delimiter)) {
-        source += "(?:" + escape(delimiter) + "$)?";
-      }
+      source += toRegExpSource(tokens, delimiter, keys, originalPath, trailing);
       combinations++;
     });
   }
@@ -484,14 +482,6 @@ export function pathToRegexp(
   pattern += end ? "$" : "(?=" + escape(delimiter) + "|$)";
 
   return { regexp: new RegExp(pattern, sensitive ? "" : "i"), keys };
-}
-
-function tokensEndWithDelimiter(
-  tokens: Exclude<Token, Group>[],
-  delimiter: string,
-): boolean {
-  const last = tokens[tokens.length - 1];
-  return last?.type === "text" && last.value.endsWith(delimiter);
 }
 
 /**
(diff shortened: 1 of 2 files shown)


src/index.ts
```diff
@@ -471,19 +471,29 @@ export function pathToRegexp(
 
       if (combinations > 0) source += "|";
       source += toRegExpSource(tokens, delimiter, keys, data.originalPath);
+      if (trailing && !tokensEndWithDelimiter(tokens, delimiter)) {
```
blakeembrey: This should occur within `toRegExpSource` to reduce the amount of code needed for this. We already know what the last token is.

blakeembrey (commented):
I'm not really convinced this is worth doing. No one has run into it as an issue, and while it might be considered a minor correctness issue, the logic of it is "simple" today. The change has to land in a major just to be sure no one is relying on it accidentally. If we can do it without any change in overall bytes I'd consider it.

iambumblehead:
@vibhor-aggr thanks; here's hoping this PR goes well and eventually gets in. Very nice that its requires only a few lines.

vibhor-aggr:
Thanks for the review. I moved the trailing delimiter handling into `toRegExpSource` in `4137f2e7e29731830da32513b072721d6abf7e7b`, so each flattened alternative decides whether it needs the optional delimiter. The current diff removes the global trailing suffix and avoids appending it when the alternative already ends with the delimiter.

Verification from the follow-up run:
- focused `/test//` Vitest: 3 passing
- full `src/index.spec.ts`: 376 passing
- `npm run build`: passed
- `npm run size`: 1.94 kB minified and brotlied, under the 2 kB limit
- `git diff --check master..HEAD`: passed

I also attempted full `npm test`; it reached 477 passing Vitest tests, but this host cannot run the ReDoS helper binary because `recheck-linux-x64` requires GLIBC_2.32/GLIBC_2.34, so that command exits with EPIPE after the loader failure.

Yek666 (approved):

iambumblehead:
This PR is strange in various ways and Blake expressed disinterest in this PR and its ticket. Perhaps "vibhor" in "vibhor-aggr" stands for something viber/vibe-coder and Blake and I are the only real people here. Maybe Blake is not "real" either I don't know anymore.

PR #444, opened 2026-06-23
fix: reject an empty-string param in compile

spokodev:
## Problem

`compile()` accepted an empty-string value for a required param, producing an empty path segment instead of rejecting it. For example, `compile("/:test")({ test: "" })` returned `/` instead of throwing.

## Fix

Reject an empty (or non-string) param value with a single check that mirrors the existing array validation, throwing `Expected "<name>" to be a non-empty string`.

## Tests

Added a case asserting `{ test: "" }` throws, and updated the one existing test that asserted the old `to be a string` message.

### DIFF
--- a/src/cases.spec.ts
+++ b/src/cases.spec.ts
@@ -271,6 +271,7 @@ export const COMPILE_TESTS: CompileTestSet[] = [
     tests: [
       { input: undefined, expected: null },
       { input: {}, expected: null },
+      { input: { test: "" }, expected: null },
       { input: { test: "123" }, expected: "/123" },
       { input: { test: "123/xyz" }, expected: "/123%2Fxyz" },
     ],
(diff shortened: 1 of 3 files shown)


blakeembrey:
None of the text here is correct, but I can see what you're trying to change. It seems simpler to just do `if (typeof value !== "string" || value.length === 0)` and change the error to `non-empty string` mirroring the array check. Do you want to make that change?

spokodev:
Done - collapsed it into the single check you suggested, mirroring the array validation: `if (typeof value !== "string" || value.length === 0)` throwing `Expected "<name>" to be a non-empty string`. Updated the one existing test that asserted the old "to be a string" message. I also rewrote the PR description, which had been pasted from a different change by mistake - that is likely the incorrect text you were looking at.

blakeembrey (approved):
While you're on this topic, would you like to work on a similar correctness fix for the wildcard? It's also technically allowing an empty string with `[""]` which wouldn't match again.

krystian2236 (commented):
he są hh

PR #447, opened 2026-07-08
Escape quoted parameter names the way parse understands

naveentehrpariya:
`stringifyName` uses `JSON.stringify` to quote non-identifier parameter names, but `parse` only recognizes a backslash as escaping the next literal character. The two schemes agree for `\"` and `\\` — and diverge for every character JSON escapes into a sequence (`\t`, `\n`, `\uXXXX`, …), silently corrupting the round-trip:

```js
const a = parse(':"a\tb"'); // name === 'a\tb' (tab is legal inside quotes)
const s = stringify(a);      // ':"a\tb"'  (JSON-escaped)
const b = parse(s);          // name === 'atb'  ← backslash stripped, 't' kept
```

This changes which parameter a compiled/matched route produces, with no error anywhere.

The fix quotes names by escaping only `"` and `\` and emitting all other characters directly, which `parse` reads back unchanged.

Also adds stringify cases for a control character and quote/backslash in names, plus a round-trip assertion that `parse(stringify(data))` reproduces the original tokens for every stringify case — that assertion fails on `master` for the new cases and passes here.

### DIFF
--- a/src/cases.spec.ts
+++ b/src/cases.spec.ts
@@ -190,6 +190,20 @@ export const STRINGIFY_TESTS: StringifyTestSet[] = [
   },
   {
     data: new TokenData([
+      { type: "text", value: "/" },
+      { type: "param", name: "a\tb" },
+    ]),
+    expected: '/:"a\tb"',
+  },
+  {
+    data: new TokenData([
+      { type: "text", value: "/" },
+      { type: "param", name: 'a"b\\c' },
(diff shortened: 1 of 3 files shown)


krzysdz:
Isn't this a duplicate of #445?

blakeembrey:
Thanks @krzysdz. I'm going to close this in favor of the first PR.

PR #450, opened 2026-07-24
Refactor matching to be non-greedy

blakeembrey:
After taking a look at https://github.com/pillarjs/path-to-regexp/pull/449 I thought I'd make another attempt at an idea I tried in the past to block backtracking, and it looks like it's working. Uses back references to reject the previous match (but only when the previous match was a param _and_ there's another wildcard).

Fixes the test case provided in that PR, and also the concerns I brought up in my comment.

### DIFF
--- a/src/index.spec.ts
+++ b/src/index.spec.ts
@@ -260,11 +260,15 @@ describe("path-to-regexp", () => {
 
   describe.each(MATCH_TESTS)(
     "match $path with $options",
-    ({ path, options, tests }) => {
-      it.each(tests)("should match $input", ({ input, expected }) => {
-        const fn = match(path, options);
-        expect(fn(input)).toEqual(expected);
-      });
+    ({ path, options, tests, testOptions }) => {
+      it.each(tests)(
+        "should match $input",
+        testOptions ?? {},
+        ({ input, expected }) => {
+          const fn = match(path, options);
+          expect(fn(input)).toEqual(expected);
+        },
+      );
     },
   );
 });
--- a/src/cases.spec.ts
+++ b/src/cases.spec.ts
@@ -1,8 +1,9 @@
+import { TestOptions } from "vitest";
+
 import {
   type MatchOptions,
   type Match,
   type ParseOptions,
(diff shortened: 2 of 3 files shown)


blakeembrey:
Actually I'm crazy and this doesn't work completely, it just blocks backtracking when it's not repeating the same captured text. Not much of an improvement.

bjohansebas (commented):
I would like to review this more carefully

blakeembrey:
I pushed up an approach that's quadratic time and I think that's the best that's possible without having some paths just fail like they do today. I don't think it's possible to get better because some level of backtracking has to be done to deal with the false positive matches in these cases.

blakeembrey:
OK, to remove quadratic time issues I pulled the idea I had from https://github.com/pillarjs/path-to-regexp/pull/426. This now matches everything as expected, but it changes _how_ it matches.

In my first attempt I only changed wildcards to make the last one the greedy one, but to fix all potential captures _and_ have it not be vulnerable to ReDoS we need to shift the parameter captures to also have the last one be greedy. The pattern this affects most is `/:path.:ext` because it becomes `[foo].[ext1.ext2]` instead of `[foo.ext1].[ext2]`.

Since it changes the behavior enough I think it should land in a new major, but I'll bring it up in the next TC meeting to ensure this is an acceptable tradeoff.

src/index.spec.ts
```diff
@@ -260,11 +260,15 @@ describe("path-to-regexp", () => {
...
     "match $path with $options",
-    ({ path, options, tests }) => {
-      it.each(tests)("should match $input", ({ input, expected }) => {
-        const fn = match(path, options);
-        expect(fn(input)).toEqual(expected);
-      });
+    ({ path, options, tests, testOptions }) => {
+      it.each(tests)(
+        "should match $input",
+        testOptions ?? {},
```
jonchurch: `testOptions` isnt used by any tests currently

Making room for the future or leftover from dev time?

UlisesGascon (approved):

PR #455, opened 2026-09-17
fix: improve regex generation for params and wildcards in segments

abhayagarwal-dev:
## Problem

On architectures like s390x, 5 tests in `redos.spec.ts` were
timing out at the 10s limit:

- `/*a@@*b~.~:c`
- `/*a-.-*b@@:c`
- `/*a@.@*b--:c`
- `/*a-@-*b..:c`
- `/x/*a/*b/y`

The root cause was that `toRegExpSource` was passing the full
`backtrack` string into `negate()` when only the first character is
meaningful as a delimiter boundary. This produced overly complex
lookahead patterns like `(?:(?!@@)[^])+` that take too long for
`recheck` to verify as safe on slower CPUs.

## Changes

- Use `backtrack[0]` instead of the full `backtrack` string when
  negating delimiters for params and wildcards in segments
- Add a deduplication check: when `wildcardBacktrack === delimiter`,
  emit `(${negate(delimiter, "")}+)` directly instead of generating
  a redundant two-alternative pattern
- Use `peekText(index)[0]` (first character only) as the second
  argument to `negate()` for the wildcard case, keeping the pattern
  minimal

## Result

All 484 tests pass on both x64/arm64 and s390x within the 10s timeout.
The generated patterns are simpler and tighter, making them faster for
`recheck` to analyse on any architecture.


blakeembrey:
This is a breaking change.

```
   Pattern       Input               Previously captured    PR result
  ━━━━━━━━━━━━  ━━━━━━━━━━━━━━━━━━  ━━━━━━━━━━━━━━━━━━━━━  ━━━━━━━━━━━
   /:a--:b       /foo--bar-baz       b: "bar-baz"           false
  ────────────  ──────────────────  ─────────────────────  ───────────
   /*a@@:b       /foo@@bar@baz       b: "bar@baz"           false
  ────────────  ──────────────────  ─────────────────────  ───────────
   /*a@@*b       /foo@@bar@baz       b: ["bar@baz"]         false
  ────────────  ──────────────────  ─────────────────────  ───────────
   /*a-*b.txt    /foo-bar.baz.txt    b: ["bar.baz"]         false
```

abhayagarwal-dev:
Thank you for the review. I will correct my code to solve timeout without failing other.

blakeembrey:
I wouldn't spend much time on this, there's a refactor of matching entirely in https://github.com/pillarjs/path-to-regexp/pull/450 which might already fix it. Additionally, tests aren't failing on CI, the work you're doing is just to work around ReDoS checks which can just be skipped locally if they're timing out.

PR #456, opened 2026-09-19
Fix stringify lookahead across empty text tokens

jakezwang:
`stringify()` can merge literal text into a parameter name when an empty text token sits between them. For custom `TokenData` containing `param("foo"), text(""), text("bar")`, it emits `:foobar` instead of `:"foo"bar`. Compiling the serialized path then rejects the original `foo` parameter.

Skip empty text tokens when looking ahead to decide whether a name needs quotes. Adds parameter and wildcard regressions with ASCII and astral identifier suffixes, plus delimiter and end-of-input controls. Each case also checks that compilation survives the stringify round trip.

Validation: `npm test` on Node 18, 20, 24, and 26; 492 tests pass with build, type checking, formatting, coverage, and the 2 kB size limit.


src/index.ts
```diff
@@ -673,9 +673,12 @@ function quoteName(name: string): string {
 /**
  * Stringify a parameter name, escaping when it cannot be emitted directly.
  */
-function stringifyName(name: string, next: Token | undefined): string {
+function stringifyName(name: string, tokens: Token[], index: number): string {
   if (!ID.test(name)) return quoteName(name);
 
+  let next = tokens[index];
+  while (next?.type === "text" && !next.value) next = tokens[++index];
```
blakeembrey: I simplified the code slightly by folding the entire thing into a `while` loop. Additionally a few bytes were shaved by removing the default value in the destructing of `token.value` now that the length is being checked.

